PR #5, opened 2019-07-16
Add cop that checks for valid sigil

peterzhu2118:
Adds a new cop `ValidSorbetSigil` that checks that every Ruby file contains a valid Sorbet sigil. Has autocorrect features that adds `typed: false` to all files that do not have a sigil. I've tested this on core and it works as expected and causes no typecheck errors.

I've talked to @clarkdave and has given me permission to adapt from this gist: https://gist.github.com/clarkdave/85aca4e16f33fd52aceb6a0a29936e52.


lib/rubocop/cop/sorbet/valid_sorbet_sigil.rb
```diff
@@ -0,0 +1,79 @@
...
+        private
+
+        def sorbet_typed_sigil_comment(processed_source)
+          processed_source.find_token do |token|
+            /#\s+typed:\s+([\w]+)/.match?(token.text)
+          end
+        end
+
+        def valid_sorbet_strictness?(strictness)
+          %w(ignore false true strict strong).include?(strictness)
```
Morriar: The is also `__STDLIB_INTERNAL` but I'm not sure it should be accepted in user code.

lib/rubocop/cop/sorbet/valid_sorbet_sigil.rb
```diff
@@ -0,0 +1,79 @@
...
+              message: 'No Sorbet sigil found in file. ' \
+                'Try a `typed: false` to start (you can also use `rubocop -a` to automatically add this).'
+            )
+          else
+            token = sorbet_typed_sigil_comment(processed_source)
+
+            add_offense(
+              token,
+              location: token.pos,
+              message: "Invalid Sorbet sigil #{strictness}."
```
Morriar: nitpick: ```Invalid Sorbet sigil `foo`.```

lib/rubocop/cop/sorbet/valid_sorbet_sigil.rb
peterzhu2118: I don't think we should encourage that to be used in anywhere except in some rbi files. It's not a problem right now, so I would prefer to not have `__STDLIB_INTERNAL` as a valid strictness.

XrXr:
Instead of searching twice, why not search the first block of comment tokens?

something like: `processed_source.tokens.take_while {|token| token.type == :tCOMMENT}.find { |token| /^ # typed: (\w+)/.match(token.text){|m| sigil = m[1] }; !!sigil }`

This way you wouldn't have to search twice to find the location of the taken

lib/rubocop/cop/sorbet/valid_sorbet_sigil.rb
```diff
@@ -0,0 +1,81 @@
...
+          end
+        end
+
+        private
+
+        SORBET_SIGIL_REGEX = /#\s+typed:\s+([\w]+)/
+
+        def sorbet_typed_sigil_comment(processed_source)
+          processed_source.tokens
+            .take_while { |token| token.type == :tCOMMENT }
```
paracycle: What if there are empty lines before the comments? I think that's still valid, no? But I assume this code will not catch it.

Morriar: Sorbet allows the `typed` sigil to be anywhere in the document.

But maybe this cop should enforce that the sigil is in the first lines (before or after the `frozen_string_literal`)

peterzhu2118: It ignores whitespace. I've added a test to demonstrate this.

paracycle: Ohhh, I see the test now. Sorry, totally missed it in my first reading, even though I looked for it. My bad, this looks good.

paracycle: Funny anecdote here: We have a test file in tapioca that starts like this:
```ruby
# frozen_string_literal: true

require "spec_helper"
require "pathname"
require "shellwords"

module Contents
  FOO_RBI = <<~CONTENTS
    # This file is autogenerated. Do not edit it by hand. Regenerate it with:
    #   generate command

    # typed: true

    module Foo
      def self.bar(a = _, b: _, **opts); end
    end

    Foo::PI = T.let(T.unsafe(nil), Float)
  CONTENTS
...
```

When I ran `srb init` the other day, it actually changed the `# typed: true` in there to `# typed: false` 😄

paracycle: Something similar might be a good test to add, btw:
```ruby

# frozen_string_literal: true

SOMETHING = <<~RUBY
  # typed: true
  Some Ruby code
RUBY
```
kind of thing. For this file, the cop should complain that it is NOT typed.

peterzhu2118: Added a test for that!

PR #6, opened 2019-07-16
Add `RequireSigilOnAllFiles` config flag to `ValidSorbetSigil` cop

peterzhu2118:
Adds a `RequireSigilOnAllFiles` config flag that when `false`, only checks for the syntax of the sigils and does not require that all files have a Sorbet sigil.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,0 +1,3 @@
+Sorbet/ValidSorbetSigil:
+  Enabled: true
+  EnforcedStyle: typed_files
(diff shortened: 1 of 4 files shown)


XrXr:
https://relishapp.com/rspec/rspec-core/v/2-5/docs/example-groups/shared-example-group#providing-context-to-a-shared-group-using-a-block maybe use a shared group to avoid repeating the autocorrect tests?

PR #8, opened 2019-09-05
Add `on_failure` to signature build order cop

ethanjgoldberg:
This cop was raising exceptions on files with `on_failure`, due to `nil` comparisons in the `sort_by` on line 44. Adding the symbol to the ORDER hash fixes this.


ethanjgoldberg:
👋 Can we get this merged?

PR #10, opened 2019-09-19
Introduce Sorbet/MandatorySorbetSigil

Morriar:
While we already have `ValidSorbetSigil` to do that, sometimes you just need to check wether a file has a sigil or not (independently of checking if the sigil is a valid one).

This pull request introduces a new cop: `MandatorySorbetSigil` used specifically to ensure that all files have a `# typed: something` at the top.

Having two different cops works better with gradual typing as you can both:
* Force the usage of a sigil in a specific set of files with `MandatorySorbetSigil`
* Check that both mandatory and non-mandatory files have a correct sigil with `ValidSorbetSigil`

Note: this PR is useless if there is a way to run `ValidSorbetSigil` with two different configs at the same time but I didn't find anything related to that in Rubocop docs.


Morriar:
@rafaelfranca closing this one in favor of #11.

PR #11, opened 2019-09-20
Add cop for each `typed` level

Morriar:
Supercede #10.

This PR introduce a cop for each Sorbet strictness level to make them mandatory:
* `IgnoreSigil`: to make `# typed: ignore` mandatory on all files
* `FalseSigil`: to make `# typed: false` mandatory on all files
* ... to `StrongLevel`

Each of them allow the autocorrect to the specified level and mark offense if the strictness level is not at least the one asked from the cop.

So a file with `typed: true`:
* will NOT make offense for cops Ignore, False and True
* will make offense with cops Strict and Strong

In addition it adds the cop `HasSigil` from #10.

And also refactor the `ValidSigil` to make subclassing easier and add some options:
* `SuggestedStrictness`: the strictness level suggested in offense messages and autocorrects
* `MinimumStrictness`: the minimum strictness level required for files (overwrite `SuggestedStrictness` if set).

Note that these options can only be used with `ValidSigil` and `HasSigil`.

See the specs for a lot of examples.


ghiculescu:
Hey Shopify team, do you know of a way to have this run at a different strictness level for different folders in a repo? Similar to how you can have multiple `.eslintrc` files in a project, and the rules in the "closest" will apply.

We have a _big_ repo, at the moment we are just using this rule to prevent `typed: ignore`. It would be great if we could say things like "everything in `app/models` must be at least `typed: strict`"  while still allowing `typed: true` in `app/controllers`.

Morriar:
You should be able to add rules in your `.rubocop.yml` like this:

```yml
Sorbet/IgnoreSigil:
  Enabled: true
  Include:
    - folder1/
    - folder2/

Sorbet/StrictSigil:
  Enabled: true
  Include:
    - folder3/
```

This will enforce a `typed: ignore` in `folder1` and `folder2` and enforce a `typed: strict` in `folder3`.

ghiculescu:
Oh of course, it's a different rule for each strictness level. That's great, thank you!

ghiculescu:
Is this going to be pushed to rubygems soon? @Morriar

Morriar:
@ghiculescu 

> Is this going to be pushed to rubygems soon? @Morriar

Haven't got the change to deploy / test it internally yet, but yes once it's stabilized.

PR #12, opened 2019-10-16
Factorize signatures related cops.

Morriar:
This PR moves all the signatures related cops to `signatures/`.

It adds an abstract cop (`SignatureCop`) that factorizes the `sig` detection and allows us to simplify the implementation of other signature related cops (see `CheckedTrueInSignature` for an example).

I didn't touch `allow_incompatible_override` as it seems to be directed at a really old version of Sorbet. Maybe we should 🔥it in another PR?


lib/rubocop/cop/sorbet/signatures/checked_true_in_signature.rb
```diff
@@ -18,13 +19,9 @@ module Sorbet
       #
       #   # good
       #   sig { void }
-      class CheckedTrueInSignature < RuboCop::Cop::Cop
+      class CheckedTrueInSignature < SignatureCop
```
paracycle: Actually does this cop even apply anymore? It seems directly related to `waffle-cone` which we are not using anymore. I guess we can 🔥 it.

lib/rubocop/cop/sorbet/signatures/signature_cop.rb
```diff
@@ -0,0 +1,28 @@
...
+        def_node_matcher(:signature?, <<~PATTERN)
+          (block (send nil? :sig) (args) ...)
+        PATTERN
+
+        def on_block(node)
+          on_signature(node) if signature?(node)
+        end
+
+        def on_signature(_)
+          # To be defined in subclasses
```
paracycle: Should we raise a `NotImplementedError` to fail fast in case someone tries to use the cop on its own?

Morriar: You could be subclassing this only to get the `signature?` node matcher without wanting to trigger something on each `sig` found.

For example [this cop](https://github.com/Shopify/rubocop-sorbet/pull/13/files#diff-f9ca2a82c8f0dbba33cb5736d1199804R34-R45), while it handles `sig` and use the matcher, don't need the `on_signature` trigger. Adding a raise `NotImplementedError` would make it throw errors when everything is perfectly fine.

lib/rubocop/cop/sorbet/signatures/checked_true_in_signature.rb
Morriar: `checked` is still in Sorbet spec: https://sorbet.org/docs/runtime#checked-whether-to-check-in-the-first-place

It uses symbols instead of `true/false`, it needs to be updated but is still relevant.

paracycle: Right, I forgot about that. Then I guess the check should be updated as well as the error message. But it can stay like this for now. We can change it in another PR.

PR #13, opened 2019-10-16
Introduce `EnforceSignatures` cop

Morriar:
This cop checks that every method definition and attribute accessor has a Sorbet signature.

It also suggest an autocorrect with placeholders so the following code:

```
def foo(a, b, c); end
```

Will be corrected as:

```
sig { params(a: T.untyped, b: T.untyped, c: T.untyped).returns(T.untyped)
def foo(a, b, c); end
```

You can configure the placeholders used by changing the following options:

* `ParameterTypePlaceholder`: placeholders used for parameter types (default: 'T.untyped')
* `ReturnTypePlaceholder`: placeholders used for return types (default: 'T.untyped')

Note: the first 3 commits belongs to #12 you should focus only on commit 3a104f2.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
+++ b/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
@@ -1,0 +1,135 @@
+# frozen_string_literal: true
+
+require 'rubocop'
+require 'stringio'
+require_relative 'signature_cop'
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop checks that every method definition and attribute accessor has a Sorbet signature.
+      #
+      # It also suggest an autocorrect with placeholders so the following code:
+      #
+      # ```
+      # def foo(a, b, c); end
+      # ```
+      #
+      # Will be corrected as:
+      #
+      # ```
+      # sig { params(a: T.untyped, b: T.untyped, c: T.untyped).returns(T.untyped)
+      # def foo(a, b, c); end
+      # ```
+      #
+      # You can configure the placeholders used by changing the following options:
+      #
+      # * `ParameterTypePlaceholder`: placeholders used for parameter types (default: 'T.untyped')
+      # * `ReturnTypePlaceholder`: placeholders used for return types (default: 'T.untyped')
+      class EnforceSignatures < SignatureCop
+        def_node_matcher(:accessor?, <<-PATTERN)
+          (send nil? {:attr_reader :attr_writer :attr_accessor} ...)
+        PATTERN
+
+        def on_def(node)
+          check_node(node)
+        end
+
+        def on_defs(node)
+          check_node(node)
+        end
+
+        def on_send(node)
+          return unless accessor?(node)
+          check_node(node)
+        end
+
+        def autocorrect(node)
+          lambda do |corrector|
+            suggest = SigSuggestion.new(node.loc.column, param_type_placeholder, return_type_placeholder)
+
+            if node.is_a?(RuboCop::AST::DefNode) # def something
+              node.arguments.each do |arg|
+                suggest.params << arg.children.first
+              end
+            elsif accessor?(node) # attr reader, writer, accessor
+              method = node.children[1]
+              symbol = node.children[2]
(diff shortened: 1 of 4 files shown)


spec/cop/sorbet/signatures/enforce_signatures_spec.rb
```diff
@@ -0,0 +1,227 @@
...
+            def foo; end
+            def bar(a, b = 2, c: Foo.new); end
+            def baz(&blk); end
+            def self.foo(a, b, &c); end
+          RUBY
+        ).to(eq(<<~RUBY))
+          sig { returns(T.untyped) }
+          def foo; end
+          sig { params(a: T.untyped, b: T.untyped, c: T.untyped).returns(T.untyped) }
+          def bar(a, b = 2, c: Foo.new); end
```
paracycle: The full list of different types of arguments is like this (from `tapioca` code):

```ruby
            when :req
              name.to_s
            when :opt
              "#{name} = _"
            when :rest
              "*#{name}"
            when :keyreq
              "#{name}:"
            when :key
              "#{name}: _"
            when :keyrest
              "**#{name}"
            when :block
              "&#{name}"
            end
```
We seem to be missing cases of splats (`*d`), kwargs (`**e`) and required keyword arguments (`f:`). Can we add them too please?

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
```diff
@@ -0,0 +1,130 @@
...
+            @param_placeholder = param_placeholder
+            @return_placeholder = return_placeholder
+          end
+
+          def to_autocorrect
+            out = StringIO.new
+            out << 'sig { '
+            unless @params.empty?
+              out << 'params('
+              @params.each_with_index do |param, index|
```
paracycle: I'd prefer:
```ruby
out << @params.map do |param|
  "#{param}: #{@param_placeholder}"
end.join(", ")
```
which is more idiomatic Ruby and does not need the `index` parameter.

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
```diff
@@ -0,0 +1,130 @@
...
+                out << "#{param}: #{@param_placeholder}"
+                if index < @params.size - 1
+                  out << ', '
+                end
+              end
+              out << ').'
+            end
+            out << if @returns.nil?
```
paracycle: It would be more readable if this block was extracted to a method and this line became:
```ruby
out << generate_return
```
or something.

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
paracycle: Also, it would be great if that was extracted to a method and this line was:
```ruby
out << generate_params
```
or something.

spec/cop/sorbet/signatures/enforce_signatures_spec.rb
Morriar: Not sure what you mean as I donc think you can express those in Sorbet syntax? See https://sorbet.org/docs/error-reference#7019 for example.

I added a test to show the behaviour of this cop nonetheless: 
https://github.com/Shopify/rubocop-sorbet/pull/13/files#diff-d7dfba5ec2240af1506fc3c02f5e3d69R123-R126

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
Morriar: Done.

It's indeed cleaner, thanks!

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
Morriar: Done too.

spec/cop/sorbet/signatures/enforce_signatures_spec.rb
paracycle: Yes, that added test was exactly what I meant. Thanks.

PR #14, opened 2019-10-24
Add a cop to enforce magic comments ordering and place the `typed` sigil at the top

Morriar:
This cop checks that the Sorbet sigil comes as the first magic comment in the file.

The expected order for magic comments is: `typed`, `(en)?coding`, `warn_indent` then `frozen_string_literal`.

For example, the following bad ordering:

```ruby
# frozen_string_literal: true
# typed: true
class Foo; end
```

Will be corrected as:

```ruby
# typed: true
# frozen_string_literal: true
class Foo; end
```

Only `typed`, `(en)?coding`, `warn_indent` and `frozen_string_literal` magic comments are considered, other comments or magic comments are left in the same place.

Signed-off-by: Alexandre Terrasa <alexandre.terrasa@shopify.com>


lib/rubocop/cop/sorbet/sigils/sigil_first.rb
```diff
@@ -0,0 +1,108 @@
...
+
+          tokens = extract_magic_comments(processed_source)
+          return if tokens.empty?
+
+          check_magic_comments_order(tokens)
+        end
+
+        def autocorrect(_node)
+          lambda do |corrector|
+            # puts node
```
rafaelfranca:
```suggestion
```

lib/rubocop/cop/sorbet/sigils/sigil_first.rb
```diff
@@ -0,0 +1,105 @@
...
+      #
+      # ```ruby
+      # # typed: true
+      # # frozen_string_literal: true
+      # class Foo; end
+      # ```
+      #
+      # Only `typed`, `(en)?coding`, `warn_indent` and `frozen_string_literal` magic comments are considered,
+      # other comments or magic comments are left in the same place.
+      class SigilFirst < ValidSigil
```
paracycle: I am not too happy with this name. Can we find a better one? This cop seems to enforce the order of sigils, so maybe something like `EnforceSigilOrder` or `EnforceSigilOrdering`?

Btw, we keep calling these sigils but I haven't found any documentation calling them sigils. Can we find the proper name for these?

Morriar: Yeah this gem needs a whole renaming / uniforming pass... I can change it to `EnforceSigilOrder`. `HasSigil` and friends should also be renamed (in another PR).

Regarding the `sigil` word, it's coming from [Sorbet official documentation](https://sorbet.org/docs/static#file-level-granularity-strictness-levels):

> However, in step (3), most kinds of errors are silenced by default, instead of being reported. To opt into more checks, we use # typed: sigils[1]. 

> Google defines sigil as, “an inscribed or painted symbol considered to have magical power,” and we like to think of types as pretty magical 🙂 ↩

I would like to keep that word as to stay coherent with the domain.

Morriar:
Reroll: changed the name of the cop to `EnforceSigilOrder`

lib/rubocop/cop/sorbet/sigils/sigil_first.rb
paracycle: Ah ok, `sigil` makes sense if Sorbet is using that term too. I'd missed that usage.

paracycle: Ah ok, `sigil` makes sense if Sorbet is using that term too. I'd missed that usage.

PR #16, opened 2019-11-11
Support for new `T.type_alias {}` syntax

### LINKED ISSUE #15: BindingConstantWithoutTypeAlias erroring
As of https://github.com/sorbet/sorbet/pull/1929, `T.type_alias` takes a block and evaluates lazily. See also https://sorbet.org/docs/type-aliases

`rubocop-sorbet` seems to not like this. `FooTwo` and `FooThree` both error with "Sorbet/BindingConstantWithoutTypeAlias: It looks like you're trying to bind a type to a constant. To do this, you must alias the type using T.type_alias."

```ruby
# typed: strict

class Bar
  extend T::Sig

  FooOne = T.type_alias(T.any(String, Integer))
  FooTwo = T.type_alias { T.any(String, Integer) }
  FooThree = T.type_alias do
    T.any(String, Integer)
  end
end
```

Morriar:
This PR updates the `BindingConstantWithoutTypeAlias` cop so it accepts the block syntax for `T.type_alias`.

I also added a custom error when the cop detects the old syntax and an autocorrect to the new block syntax so:

```ruby
Foo = T.type_alias(T.any(String, Integer))
```

Is autocorrected to:

```ruby
Foo = T.type_alias { T.any(String, Integer) }
```

Fixes #15.

cc. @ghiculescu.

### DIFF
--- a/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
+++ b/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
@@ -14,18 +14,18 @@ module RuboCop
       #   FooOrBar = T.any(Foo, Bar)
       #
       #   # good
-      #   FooOrBar = T.type_alias(T.any(Foo, Bar))
+      #   FooOrBar = T.type_alias { T.any(Foo, Bar) }
       class BindingConstantWithoutTypeAlias < RuboCop::Cop::Cop
         def_node_matcher(:binding_unaliased_type?, <<-PATTERN)
           (casgn _ _ [#not_nil? #not_t_let? #method_needing_aliasing_on_t?])
         PATTERN
 
         def_node_matcher(:using_type_alias?, <<-PATTERN)
-          (
-            send
-            (const nil? :T)
-            :type_alias
-            _
+          (block
+            (send
+              (const nil? :T) :type_alias)
+              _
+              _
           )
         PATTERN
 
@@ -69,7 +69,7 @@ module RuboCop
           lambda do |corrector|
             corrector.replace(
               node.source_range,
-              "T.type_alias(#{node.source})"
+              "T.type_alias { #{node.source} }"
             )
           end
         end
--- a/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
+++ b/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
@@ -29,6 +29,15 @@ module RuboCop
           )
         PATTERN
 
+        def_node_matcher(:using_deprecated_type_alias_syntax?, <<-PATTERN)
+          (
+            send
+            (const nil? :T)
+            :type_alias
+            _
+          )
+        PATTERN
+
         def_node_matcher(:t_let?, <<-PATTERN)
           (
             send
@@ -58,6 +67,14 @@ module RuboCop
 
         def on_casgn(node)
           return unless binding_unaliased_type?(node) && !using_type_alias?(node.children[2])
+          if using_deprecated_type_alias_syntax?(node.children[2])
+            add_offense(
+              node.children[2],
+              message: "It looks like you're using the old `T.type_alias` syntax. " \
+              '`T.type_alias` now expects a block.'
+            )
+            return
+          end
           add_offense(
             node.children[2],
             message: "It looks like you're trying to bind a type to a constant. " \
@@ -67,6 +84,14 @@ module RuboCop
 
         def autocorrect(node)
           lambda do |corrector|
+            if using_deprecated_type_alias_syntax?(node)
+              corrector.replace(
+                node.source_range,
+                "T.type_alias { #{node.children[2].source} }"
+              )
+              return
+            end
+
             corrector.replace(
               node.source_range,
               "T.type_alias { #{node.source} }"
--- a/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
+++ b/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
@@ -71,7 +71,9 @@ module RuboCop
             add_offense(
               node.children[2],
               message: "It looks like you're using the old `T.type_alias` syntax. " \
-              '`T.type_alias` now expects a block.'
+              '`T.type_alias` now expects a block.' \
+              'Run Sorbet with the options "--autocorrect --error-white-list=5043" ' \
+              'to automatically upgrade to the new syntax.'
             )
             return
           end
--- a/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
+++ b/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
@@ -84,14 +84,6 @@ module RuboCop
 
         def autocorrect(node)
           lambda do |corrector|
-            if using_deprecated_type_alias_syntax?(node)
-              corrector.replace(
-                node.source_range,
-                "T.type_alias { #{node.children[2].source} }"
-              )
-              return
-            end
-
             corrector.replace(
               node.source_range,
               "T.type_alias { #{node.source} }"
--- a/spec/cop/sorbet/binding_constant_without_type_alias_spec.rb
+++ b/spec/cop/sorbet/binding_constant_without_type_alias_spec.rb
@@ -14,7 +14,9 @@ RSpec.describe(RuboCop::Cop::Sorbet::BindingConstantWithoutTypeAlias, :config) d
 
   def deprecation
     "It looks like you're using the old `T.type_alias` syntax. " \
-    '`T.type_alias` now expects a block.'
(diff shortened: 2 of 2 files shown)


lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
```diff
@@ -67,9 +84,17 @@ def on_casgn(node)
 
         def autocorrect(node)
           lambda do |corrector|
+            if using_deprecated_type_alias_syntax?(node)
+              corrector.replace(
```
XrXr: Sorbet itself has an autocorrect for this that picks the `do; end` syntax for multi-line aliases. Maybe we can leave this to Sorbet?

Morriar: I was wondering the same thing. I wanted to have a custom error for the old syntax and while at it I was able to give the autocorrect.

I can remove the autocorrect but I would like to keep the custom message as the current want is a bit misleading (see #15).

ghiculescu: I think keeping an autocorrect here is a good idea. Sorbet doesn't let you selectively autocorrect (as far as I know), and in my experience running `srb tc -a` sometimes gets false positives. Keeping this will allow you to autocorrect just this one issue.

rafaelfranca:
> It would be super cool if we could autocorrect multi-line calls to `do`/`end` blocks, but I am content even without that.

I'm sure https://rubocop.readthedocs.io/en/stable/cops_style/#styleblockdelimiters can complement our cop.

Morriar:
@paracycle are you ok with @rafaelfranca solution for `do`/`end` blocks?

Morriar:
@paracycle good catch!

Unfortunately, I'm stuck on something else this week so I can either push this again without the auto-correct or wait for a week or so, so I can look into fixing this.

ghiculescu:
@paracycle that code isn't flagged as an error when I run this version of `rubocop-sorbet`, so  I don't think it would cause a problem here.

My test file:

```ruby
# typed: false
Foo = T.type_alias(foo: Integer, bar: String)
FooTwo = T.type_alias(T.any(foo: Integer, bar: String))
Bar = T.type_alias(T.any(String, Integer))
```

rubocop output:

```
app/test.rb:3:10: C: Sorbet/BindingConstantWithoutTypeAlias: It looks like you're using the old T.type_alias syntax. T.type_alias now expects a block.
FooTwo = T.type_alias(T.any(foo: Integer, bar: String))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/test.rb:4:7: C: Sorbet/BindingConstantWithoutTypeAlias: It looks like you're using the old T.type_alias syntax. T.type_alias now expects a block.
Bar = T.type_alias(T.any(String, Integer))
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Note how your sample line wasn't flagged as an error.

sorbet output:

```
app/test.rb:2: No block given to T.type_alias https://srb.help/5043
     2 |Foo = T.type_alias(foo: Integer, bar: String)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    app/test.rb:2: Replace with T.type_alias {{foo: Integer, bar: String}}
     2 |Foo = T.type_alias(foo: Integer, bar: String)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

app/test.rb:3: No block given to T.type_alias https://srb.help/5043
     3 |FooTwo = T.type_alias(T.any(foo: Integer, bar: String))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    app/test.rb:3: Replace with T.type_alias {T.any(foo: Integer, bar: String)}
     3 |FooTwo = T.type_alias(T.any(foo: Integer, bar: String))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

app/test.rb:4: No block given to T.type_alias https://srb.help/5043
     4 |Bar = T.type_alias(T.any(String, Integer))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  Autocorrect: Use `-a` to autocorrect
    app/test.rb:4: Replace with T.type_alias {T.any(String, Integer)}
     4 |Bar = T.type_alias(T.any(String, Integer))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

test file after autocorrecting with rubocop:

```ruby
# typed: false
Foo = T.type_alias(foo: Integer, bar: String)
FooTwo = T.type_alias { T.any(foo: Integer, bar: String) }
Bar = T.type_alias { T.any(String, Integer) }
```

Rubocop hasn't changed your sample line.

test file file after autocorrecting with sorbet:

```ruby
# typed: false
Foo = T.type_alias {{foo: Integer, bar: String}}
FooTwo = T.type_alias { T.any(foo: Integer, bar: String) }
Bar = T.type_alias { T.any(String, Integer) }
```

This appears to match the desired behaviour.

paracycle:
> @paracycle that code isn't flagged as an error when I run this version of `rubocop-sorbet`, so I don't think it would cause a problem here. Let me know if I'm testing this wrong.

Apologies for the super late response on this. Commenting inline below.

> test file after autocorrecting with rubocop:
> 
> ```ruby
> # typed: false
> Foo = T.type_alias(foo: Integer, bar: String)
> FooTwo = T.type_alias { T.any(foo: Integer, bar: String) }
> Bar = T.type_alias { T.any(String, Integer) }
> ```
> 
> Rubocop hasn't changed your sample line.

And indeed, the resulting code is wrong, since `T.type_alias(foo: Integer, bar: String)` will not be valid Sorbet code, it should have been corrected to `T.type_alias{ {foo: Integer, bar: String} }`.

> 
> test file file after autocorrecting with sorbet:
> 
> ```ruby
> # typed: false
> Foo = T.type_alias {{foo: Integer, bar: String}}
> FooTwo = T.type_alias { T.any(foo: Integer, bar: String) }
> Bar = T.type_alias { T.any(String, Integer) }
> ```
> 
> This appears to match the desired behaviour.

Indeed, Sorbet does the right correction here.

Basically my point is the correction that this Rubocop rule attempts to make is mostly correct but fails in some specific ways, whereas the Sorbet autocorrection is always correct.

With that in mind, I'd ultimately agree with @XrXr that we should probably merge this WITHOUT the autocorrect in place. We could look into adding the autocorrect later.

Morriar:
Ok, I removed the autocorrect.

lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
```diff
@@ -69,7 +86,7 @@ def autocorrect(node)
           lambda do |corrector|
             corrector.replace(
               node.source_range,
-              "T.type_alias(#{node.source})"
+              "T.type_alias { #{node.source} }"
```
paracycle: Is this still not attempting to autocorrect?

Morriar: I kept the original behavior nesting 

```ruby
# bad
FooOrBar = T.any(Foo, Bar)
```

in:

```ruby
# good
FooOrBar = T.type_alias { T.any(Foo, Bar) }
```

What I removed is:

```ruby
if using_deprecated_type_alias_syntax?(node)
  corrector.replace(
    node.source_range,
    "T.type_alias { #{node.children[2].source} }"
  )
  return
end
```

paracycle: Oh, right. Sorry, I should have read the code better. Makes sense.

lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb
```diff
@@ -58,6 +67,14 @@ def not_nil?(node)
 
         def on_casgn(node)
           return unless binding_unaliased_type?(node) && !using_type_alias?(node.children[2])
+          if using_deprecated_type_alias_syntax?(node.children[2])
+            add_offense(
+              node.children[2],
+              message: "It looks like you're using the old `T.type_alias` syntax. " \
+              '`T.type_alias` now expects a block.'
```
paracycle: Should we add a note here on how to fix this? For example:
`Run Sorbet with the options "--autocorrect --error-white-list=5043" to automatically upgrade to the new syntax.`

Morriar: ^ I like this, pushed a new version with the error message.

PR #17, opened 2019-11-25
Accept splats after default keyword parameters

Morriar:
As raised by @Mangara:

This snippet should not raise any error:

```ruby
sig { params(a: String, b: Integer, c: String, d: Integer).void }
def foo(a, b: 1, c: 'a', **d); end
```

This PR fixes the `KeywordArgumentOrdering` cop so those constructs are accepted (see [`tests`](https://github.com/Shopify/rubocop-sorbet/commit/0dd142436efeb72801f820b38c8be9eae10d5471#diff-fd5be9f30e2845592e2bfae314bfd456) for more details).

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/keyword_argument_ordering.rb
+++ b/lib/rubocop/cop/sorbet/signatures/keyword_argument_ordering.rb
@@ -35,7 +35,7 @@ module RuboCop
           out_of_kwoptarg = false
 
           parameters.reverse.each do |param|
-            out_of_kwoptarg = true unless param.type == :kwoptarg || param.type == :blockarg
+            out_of_kwoptarg = true unless param.type == :kwoptarg || param.type == :blockarg || param.type == :kwrestarg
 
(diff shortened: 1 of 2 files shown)


Mangara:
This will allow optional keyword arguments after a splat. Is that a problem, or are there other mechanism that catch this?

Should we add a test or two for failure cases that involve a splat?

Morriar:
> This will allow optional keyword arguments after a splat. Is that a problem, or are there other mechanism that catch this?

It seems like it's already a Ruby error:

```ruby
def foo(**a, b:); end
```

Gives the following error:

```
1: syntax error, unexpected tLABEL, expecting & or '&'
def foo(**a, b:); end
             ^~
```

I think the purpose of this cop is to avoid the valid Ruby constructs not supported by Sorbet.

Morriar:
@Mangara added test: https://github.com/Shopify/rubocop-sorbet/pull/17/commits/cb288cf40e297e34a0dd38882b87964c00babbfc

PR #19, opened 2019-12-03
Use the higher of `Suggested` and `Minimum` strictness when autocorrecting

ghiculescu:
We want to use config like this:

```yml
Sorbet/FalseSigil:
  Enabled: true
  SuggestedStrictness: true
```

So that we can require at least a `false` sigil on *all* files, but add `typed: true` to new files. This way we know that Sorbet adoption will go up over time as new files are added (we have a separate Danger rule that warns against downgrading a file's type).

This PR adds the ability to do that, by choosing the higher level out of the suggested and minimum levels when autocorrecting. There is a special case when you are using `ignore` as your minimum level in which this doesn't take place, I did this rather than change existing specs so as to not create a breaking change.

### DIFF
--- a/lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
+++ b/lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
@@ -36,7 +36,8 @@ module RuboCop
             return unless extract_sigil(processed_source).nil?
 
             token = processed_source.tokens.first
-            corrector.insert_before(token.pos, "# typed: #{minimum_strictness || suggested_strictness}\n")
+            replace_with = suggested_strictness_level(minimum_strictness, suggested_strictness)
+            corrector.insert_before(token.pos, "# typed: #{replace_with}\n")
           end
         end
 
@@ -64,7 +65,7 @@ module RuboCop
 
           token = processed_source.tokens.first
           if require_sigil_on_all_files?
-            strictness = minimum_strictness || suggested_strictness
+            strictness = suggested_strictness_level(minimum_strictness, suggested_strictness)
             add_offense(
               token,
               location: token.pos,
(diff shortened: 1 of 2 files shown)


ghiculescu:
@Morriar is this okay to merge?

lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
```diff
@@ -75,6 +76,25 @@ def check_sigil_present(sigil)
...
+          # special case: if you're using Sorbet/IgnoreSigil without config, we should recommend `ignore`
+          return "ignore" if minimum_strictness == "ignore" && cop_config['SuggestedStrictness'].nil?
+
+          # if a minimum strictness is set (eg. you're using Sorbet/FalseSigil)
+          # we want to compare the minimum strictness and suggested strictness. this is because
+          # the suggested strictness might be higher than the minimum (eg. if you want all new files
+          # at a higher strictness level, without having to migrate existing files at lower levels).
+
+          suggested_level = STRICTNESS_LEVELS.index(suggested_strictness)
+          minimum_level = STRICTNESS_LEVELS.index(minimum_strictness)
```
paracycle: Can we do sanity check on these levels? Since `suggested_strictness` and `minimum_strictness` are user supplied values, there is a chance that they might be wrong, so there is a chance that these values end up being `-1`

ghiculescu: https://github.com/Shopify/rubocop-sorbet/pull/19/commits/d18620e60f7d5857dbebc348dadd25aa033f9f36

Morriar:
Thanks you @ghiculescu!

ghiculescu:
@Morriar could I ask for a new gem release please?

Morriar:
@ghiculescu it's done: https://rubygems.org/gems/rubocop-sorbet/versions/0.3.5

Thanks for the reminder :)

PR #23, opened 2020-02-07
Autocorrect adds sigil after shebang line

omarkhan:
When the Sorbet/HasSigil cop is enabled, autocorrecting causes the sigil to be added _before_ the shebang line:

```ruby
# typed: false
#!/usr/bin/env ruby
```

This prevents the script from being interpreted as a ruby script when called from the command line. This pull request checks for a shebang line and, if found, adds the sorbet sigil after it instead.


Morriar:
Released as v0.3.7: https://rubygems.org/gems/rubocop-sorbet/versions/0.3.7

PR #24, opened 2020-03-24
Enforce that the encoding sigil is the first one

casperisfine:
Otherwise it's ignored by Ruby:

```ruby
# encoding: US-ASCII
# typed: false

p "foo".encoding
```

```
$ ruby /tmp/encoding.rb
#<Encoding:US-ASCII>
```

```ruby
# typed: false
# encoding: US-ASCII

p "foo".encoding
```

```
$ ruby /tmp/encoding.rb
#<Encoding:UTF-8>
```


Morriar:
TIL, thanks!

PR #29, opened 2020-06-03
Standardize the gem layout

paracycle:
Browsing through the Rubocop documentation I noticed that it says:
> If you’d like to create an extension gem, you can use [rubocop-extension-generator](https://github.com/rubocop-hq/rubocop-extension-generator).

So, I checked `rubocop-rails` as an example and noticed that it has the same folder structure as generated by that generator. So I ran the generator on our codebase and committed the changes.

The new layout opens up a few opportunities:

1. It allows us to export our default config for the `Sorbet` department.
2. It allows us to integrate automatic documentation of our cops by pulling in a rake task.

I also took the opportunity to run gem tests in Ruby versions 2.5, 2.6 and 2.7 on CI and also brought our rubocop-shopify dependency up-to-date.


paracycle:
To be fair, the doc addition can be a totally separate PR. Let me know if you think so too.

rubocop-sorbet.gemspec
```diff
@@ -1,31 +1,31 @@
 # frozen_string_literal: true
-
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require_relative "lib/rubocop/sorbet/version"
```
peterzhu2118, line 2: Minor nit: I think we should consistently use either single quotes or double quotes. IIRC the `rubocop-shopify`  gem does not have a preference.

paracycle: Indeed and that's why I'd changed all usages in `.gemspec` to use double-quotes. It should be consistent. Am I missing something?

peterzhu2118: I think there's a mix of the two in places like [`here`](https://github.com/Shopify/rubocop-sorbet/blob/5c717f454f867dfa06b7ea9b4518169ddbb0f2ea/Gemfile#L10), [`here`](https://github.com/Shopify/rubocop-sorbet/blob/5c717f454f867dfa06b7ea9b4518169ddbb0f2ea/spec/spec_helper.rb#L3), or [`here`](https://github.com/Shopify/rubocop-sorbet/blob/5c717f454f867dfa06b7ea9b4518169ddbb0f2ea/lib/rubocop/cop/sorbet_cops.rb#L2).

paracycle: Ok, resolved some of them, but I am not going to change it all across the codebase if usages within a single file are consistent. For example, your last example.

peterzhu2118: :+1: Thanks!

Issue #35, opened 2020-06-11
Feature Request: Style/MutableConstant applied to T.let argument


dduugg:
The `Style/MutableConstant` [cop](https://docs.rubocop.org/rubocop/cops_style.html#stylemutableconstant) ensures that literal constants (or all constants in `strict` style) are frozen.

However, in a `typed: strict` file, constants must have `T.let` annotations. Sorbet does not recognize `T.let` calls that have`.freeze` appended to them. This largely invalidates the `MutableConstant` cop (since literal constants are wrapped by `T.let` calls, and the latter cannot be frozen).

I would like a sorbet-compatible version of the `Style/MutableConstant` cop. It should require literal first arguments to `T.let` to be frozen with `EnforcedStyle` set to `literals` (the default), or all first arguments when set to `strict`.

Would the maintainers be amenable to such a cop?

Morriar:
👋 Hey @dduugg,

I've been looking at #36 and I think it might not be in the scope of this project.

Our goal with `rubocop-sorbet` is to check that the code is typable and to provide the required autocorrections to make it so. In this particular case I believe the cop is more related to styling than typing.

Also, since `Style/MutableConstant` is a quite popular cop, already used by a lot of people. I think the best course of action would be to either:

1. change Sorbet behaviour so it accepts `Y = T.let(x, X).freeze`
2. change Style/MutableConstant behaviour to it accepts `Y = T.let(x.freeze, X)`

What do you think about bringing this issue to the https://github.com/sorbet/sorbet repository?

PR #36, opened 2020-06-17
Add cop for mutable T.let constants

dduugg:
Implements https://github.com/Shopify/rubocop-sorbet/issues/35

This is mostly a copy pasta of https://github.com/rubocop-hq/rubocop/blob/v0.85.1/lib/rubocop/cop/style/mutable_constant.rb (differing by adding a node matcher and making slight changes to the `on_casgn` and `on_or_asgn` impls)

The output of `bundle exec rake new_cop` requests adding an entry to `CHANGELOG.md`, which did not exist. I have taken the liberty of creating one in a second commit, modeled on https://github.com/rubocop-hq/rubocop/blob/v0.85.1/CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Change log
+
+## master (unreleased)
+
+### New features
+
+* [#35](https://github.com/Shopify/rubocop-sorbet/issues/35): Add cop for mutable T.let constants. ([@dduugg][])
+
+[@dduugg]: https://github.com/dduugg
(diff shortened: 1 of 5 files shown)


dduugg:
Aww bummer, feel free to close out if that's the case. (This is outside the scope of Rubocop core, which requires library support in modules such as rubocop-rails or rubocop-sorbet. I've chatted informally with the sorbet folks and they're disinclined to support transformations on `T.let` in strict mode.)

Thanks for your consideration though.

PR #39, opened 2020-07-22
Bump rake to version >= 12.3

Morriar:
Fixing the vulnerability with Rake, see https://github.com/advisories/GHSA-jppv-gw3r-w3q8.


Morriar:
> I don't think this actually bumped the rake version above 12.3.2 which is our current version.

Apparently I don't know how to copy/paste a line to the end 😂 . Adding the proper constraint actually bumped the version. Thanks for spotting this.

Issue #40, opened 2020-08-17
SignatureBuildOrder should preserve generics formatting


Morriar:
Right now, applying a suggestion about signatures build order will change the formatting of Sorbet's generics to use the `.` notation.

For example,
```ruby
sig { void.abstract.params(x: T::Array[String]) }
```

Is auto-corrected into:
```ruby
sig { abstract.params(x: T::Array.[](String)).void }
```

instead of:
```ruby
sig { abstract.params(x: T::Array[String]).void }
```

I believe this occurs because we use `unparser` to [transform the s expression back into a Ruby code String](https://github.com/Shopify/rubocop-sorbet/blob/master/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb#L66).

PR #41, opened 2020-08-18
Preserve generic formatting by using index send nodes

paracycle:
Fixes: #40

By default RuboCop does not enable the more modern AST syntax that users of whitequark parser can opt-in to by turning on various `emit_xxx` flags on `Parser::Builders::Default`. One of those modern Ruby AST syntax is the `index` nodes that can be emitted for index accessors, which is necessary for us to properly roundtrip generic types back to expected Ruby code.

The "fix" is an ugly hack that re-parses the given node source with `emit_index` turned on and applies `unparse` on the returned node set instead of the original one.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
@@ -55,10 +55,9 @@ module RuboCop
           return nil unless can_autocorrect?
 
           lambda do |corrector|
-            rich_node = node_with_index_sends(node)
-            nodes = call_chain(rich_node).sort_by { |call| ORDER[call.method_name] }
-            tree =
-              nodes.reduce(nil) do |receiver, caller|
+            tree = call_chain(node_with_index_sends(node))
+              .sort_by { |call| ORDER[call.method_name] }
+              .reduce(nil) do |receiver, caller|
                 caller.updated(nil, [receiver] + caller.children.drop(1))
               end
 
--- a/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
@@ -55,7 +55,8 @@ module RuboCop
           return nil unless can_autocorrect?
 
           lambda do |corrector|
-            nodes = call_chain(node).sort_by { |call| ORDER[call.method_name] }
+            rich_node = node_with_index_sends(node)
+            nodes = call_chain(rich_node).sort_by { |call| ORDER[call.method_name] }
             tree =
               nodes.reduce(nil) do |receiver, caller|
                 caller.updated(nil, [receiver] + caller.children.drop(1))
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -70,6 +70,16 @@ def autocorrect(node)
 
         private
 
+        def node_with_index_sends(node)
+          # This is really dirty hack to reparse the current node with index send
+          # emitting enabled, which is necessary to unparse them back as index accessors.
+          emit_index_value = RuboCop::AST::Builder.emit_index
+          RuboCop::AST::Builder.emit_index = true
+          RuboCop::AST::ProcessedSource.new(node.source, target_ruby_version, processed_source.path).ast
```
mbj, line 78: Instead of setting a global flag and use rubocop you can use `Unparser.parse` directly on the `node.source`. It runs (without setting global flags) the correct "modern" parser.

paracycle: We have been trying to not get a hard dependency on `Unparser` for the cop to work, but only enabled autocorrect if it existed. It seems that I was too hung up on that requirement and missed the fact that this is a correct phase operation so can depend on the existence of `Unparser`.

Thanks, I'll take a look at that.

mbj: @paracycle Cool. There is also `Unparser.parser` which returns a "correctly" configured parser, possibly you can push that one into rubocop, so rubocop still controls the parsing process.

Also I could see myself actually supporting target ruby versions in unparser, but that would require funding. For `mutant` (which is the primary reason unparser exists) this is not required.

OT: Unparser recently got all "unique" dependencies removed, see https://rubygems.org/gems/unparser/versions/0.6.0, you may be able to reduce end user usage complexity in going for a hard dependency.

I'ts likely I can remove the `diff-lcs` dependency soon in factoring out `unparser-cli` from `unparser`. But this should not too much be an issue as `diff-lcs` is a very common dependency.

paracycle: Hmm, I can't quite get `Unparser.parser` to work, since Rubocop AST works by supplying a custom `Builder` to the parser object that creates special node types with extra methods and attributes. The call to `root_call`, which is a node matcher, fails if the supplied node is not a Rubocop AST node type. :(

mbj: @paracycle I see, I wounder if you can instead supply a builder to rubocop that is defined like this? https://github.com/mbj/unparser/blob/3a459c8fbe117d3a78c1efe48c76ddfee175816f/lib/unparser.rb#L22-L31

But inheriting the rubocop builder? This way you do not set globals.

OT: I wounder what it would take to rubocop to use the modern AST format by default? You are de-facto sending it modern AST already, so the change rubocop side may be easier.

paracycle: 🤯 Of course, that seems to work beautifully: https://github.com/Shopify/rubocop-sorbet/pull/51

PR #42, opened 2020-09-03
EnforceSigilOrder should remove blank lines while reordering

KaanOzkan:
The autocorrect for this cop simply replaces magic comment's. Therefore given an input:
```
1 # frozen_string_literal: true
2
3 # typed: true
```
Autocorrect outputs:
```
1 # typed: true
2
3 # frozen_string_literal: true
```
(Line numbers are shown for easier understanding)

This PR removes empty lines **in between magic comments**

resolves https://github.com/Shopify/sorbet-issues/issues/95


lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
```diff
@@ -99,6 +108,30 @@ def check_magic_comments_order(tokens)
...
+
+            prev_line = cur_line
+          end
+        end
+
+        private
+
+        def previous_and_current_lines_empty?(line)
+          processed_source[line - 2].empty? && processed_source[line - 1].empty?
+        end
```
KaanOzkan, line 134: These 2 helpers methods were taken from https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/layout/empty_lines.rb and adapted for our use case.

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
```diff
@@ -99,6 +108,30 @@ def check_magic_comments_order(tokens)
...
+              end
+            end
+
+            prev_line = cur_line
+          end
+        end
+
+        private
+
+        def previous_and_current_lines_empty?(line)
```
paracycle: Where is this used?

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
```diff
@@ -28,6 +28,8 @@ module Sorbet
       # Only `typed`, `(en)?coding`, `warn_indent` and `frozen_string_literal` magic comments are considered,
       # other comments or magic comments are left in the same place.
       class EnforceSigilOrder < ValidSigil
+        include RangeHelp
```
paracycle, line 31: Where is this coming from and what does it do?

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
```diff
@@ -49,6 +51,13 @@ def autocorrect(_node)
             tokens.each_with_index do |token, index|
               corrector.replace(token.pos, expected[index].text)
             end
+
+            # Remove blank lines between the magic comments
+            lines = Set.new
+            tokens.each { |token| lines << token.line }
+            each_extra_empty_line(lines.sort) do |range|
+              corrector.remove(range)
+            end
```
paracycle, line 60: Can we not just `lines.each` and in the block just remove the source range for a line if it is `empty?` I am not sure if I understand what the double loop in the `each_extra_empty_line` is doing.

spec/rubocop/cop/sorbet/sigils/enforce_sigil_order_spec.rb
```diff
@@ -143,5 +143,26 @@ class Foo; end
...
+        class Foo; end
+      RUBY
+      expect(autocorrect_source(source))
+        .to(eq(<<~RUBY))
+          # encoding: utf-8
+          # typed: true
+          # frozen_string_literal: true
+          class Foo; end
+        RUBY
+    end
```
paracycle, line 166: How does it behave with:
```ruby
# frozen_string_literal: true


# typed: true

# encoding: utf-8


class Foo; end
```

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
KaanOzkan: I'm in the process of simplifying what I copy pasted over 😄  
I will try that too.

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
KaanOzkan: It gives us access to the `source_range` method https://github.com/rubocop-hq/rubocop/blob/v0.85.0/lib/rubocop/cop/mixin/range_help.rb#L11

We need a range for`corrector.remove`

spec/rubocop/cop/sorbet/sigils/enforce_sigil_order_spec.rb
KaanOzkan: As expected. I can add a test for it if you want more cases.

paracycle: Yes, that would be great. With these types of manipulations, the first mistake leads to out of bounds access and the removal spills to the surrounding area. So, it would be good to have checks around the affected region too.

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
KaanOzkan: The `lines` does not represent the lines in the file. It only consists of lines that are not empty. For example if we have 
```
1 # frozen_string_literal: true
2
3
4 # typed: true
```
`lines` contains `[1, 4]`

Therefore, in order to check for emptiness of lines `2, 3` we need to keep track of previous and current lines. The inner loop is iterating between "the first potentially empty line" and "current line". 

In our example we'd have only 1 iteration and these would be "2" and "4" respectively. Since it's iterating between a range of these, `2..4`, it's iterating over both of the potentially empty lines and checks for their emptiness.

Lmk if that example helps.

spec/rubocop/cop/sorbet/sigils/enforce_sigil_order_spec.rb
KaanOzkan: I was mistaken, it's not as expected in this case. 

It preserves any blank lines after magic comments. I've added the test below to make it clear. Note how blank lines around `class Foo; end` remain same.

paracycle: Yes, that's exactly what I would have expected. Thanks for the test.

KaanOzkan: Thank you for pointing it out. Should've been tested in the first place.

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
paracycle: Yes, but that still feels like a lot of overhead. I have a suggestion for a simpler implementation. Will resolve this and comment that.

paracycle, line 60: The change below gives me the same output. Basically, we iterate on the whole range from the smallest to the largest line number and skip the line numbers that we have sigils on. Similarly, we skip lines that are not really empty and remove all the rest.

I feel like it is much simpler to reason about.

```suggestion
            lines = tokens.map(&:line).to_set
            (lines.min...lines.max).each do |line|
              next if lines.include?(line)
              next unless processed_source[line - 1].empty?
              corrector.remove(source_range(processed_source.buffer, line, 0))
            end
```

spec/rubocop/cop/sorbet/sigils/enforce_sigil_order_spec.rb
```diff
@@ -143,5 +143,50 @@ class Foo; end
...
         RUBY
     end
+
+    it('autocorrects magic comments while removing blank lines') do
+      source = <<~RUBY
+        # frozen_string_literal: true
+
+
+        # typed: true
+
```
paracycle, line 153: Let's make sure non empty comment lines are not deleted:

```suggestion
		#
        # typed: true
		#
```

spec/rubocop/cop/sorbet/sigils/enforce_sigil_order_spec.rb
```diff
@@ -143,5 +143,50 @@ class Foo; end
...
+
+        # encoding: utf-8
+        class Foo; end
+      RUBY
+      expect(autocorrect_source(source))
+        .to(eq(<<~RUBY))
+          # encoding: utf-8
+          # typed: true
+          # frozen_string_literal: true
```
paracycle, line 163:
```suggestion
          # encoding: utf-8
          #
          # typed: true
          #
          # frozen_string_literal: true
```

KaanOzkan: I can turn these 2 suggestions into a new test?

paracycle: Sure, that would be nice.

lib/rubocop/cop/sorbet/sigils/enforce_sigil_order.rb
KaanOzkan: Hmm I think it's definitely easier to reason about. Although it's less performant right. `include?` call would make it `O(n^2)`?

Not that performance really matters since it's just a rubocop autocorrect.

paracycle: But `lines` is a `Set` so the include check should be `O(1)`.

KaanOzkan: Ah right. Pretty much same then other than iterating over all of the lines. Changing it now, thanks.

paracycle (approved):

Morriar (approved):

PR #43, opened 2020-09-16
Disable ForbidIncludeConstLiteral & ForbidSuperClassConstLiteral by default

KaanOzkan:
Since the 0.4.0 release we have a set of cops that are enabled by default. Upon a failure in shopify core we noticed that the autocorrect is also faulty. I will remove it's autocorrect in another PR. This one just disables the cops by default, they are "aggressive" and will cause failures for people. Makes sense to tell users to "opt in" rather than "opt out".

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -55,12 +55,12 @@ Sorbet/FalseSigil:
 
 Sorbet/ForbidIncludeConstLiteral:
   Description: 'Forbids include of non-literal constants.'
-  Enabled: true
+  Enabled: false
   VersionAdded: 0.2.0
 
 Sorbet/ForbidSuperclassConstLiteral:
   Description: 'Forbid superclasses which are non-literal constants.'
-  Enabled: true
+  Enabled: false
   VersionAdded: 0.2.0
 
 Sorbet/ForbidUntypedStructProps:
--- a/config/default.yml
+++ b/config/default.yml
@@ -57,11 +57,13 @@ Sorbet/ForbidIncludeConstLiteral:
   Description: 'Forbids include of non-literal constants.'
   Enabled: false
   VersionAdded: 0.2.0
+  VersionChanged: 0.5.0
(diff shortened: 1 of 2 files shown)


config/default.yml
```diff
@@ -55,12 +55,12 @@ Sorbet/FalseSigil:
 
 Sorbet/ForbidIncludeConstLiteral:
   Description: 'Forbids include of non-literal constants.'
-  Enabled: true
+  Enabled: false
```
katayounHillier, line 58: question: would it work if we just delete this line? or do we need to specifically set it to `false`

katayounHillier (approved):

config/default.yml
KaanOzkan: I couldn't find any documentation regarding the default value. Looks like even built in cops always specify `Enabled:` config https://github.com/rubocop-hq/rubocop/blob/master/config/default.yml so it maybe a faulty config to not use set this property 🤷

config/default.yml
```diff
@@ -55,12 +55,12 @@ Sorbet/FalseSigil:
...
   VersionAdded: 0.2.0
 
 Sorbet/ForbidSuperclassConstLiteral:
   Description: 'Forbid superclasses which are non-literal constants.'
-  Enabled: true
+  Enabled: false
   VersionAdded: 0.2.0
```
paracycle, line 64: We haven't been using them so far, but can we also add a `VersionChanged: 0.5.0` for these two changes?

KaanOzkan: I think it makes sense to do that in the `autocorrect` PR? Does this count as a change to the cop itself?

paracycle: According to https://github.com/rubocop-hq/rubocop/issues/5978, the defaults changing is also captured with `VersionChanged`

paracycle (approved):

PR #44, opened 2020-09-16
Remove autocorrects for ForbidSuperclassConstLiteral & ForbidIncludeConstLiteral

KaanOzkan:
Remove autocorrects for ForbidSuperclassConstLiteral & ForbidIncludeConstLiteral. They were deemed to be faulty autocorrects.

Removed the `autocorrect` methods, removed autocorrect tests, updated the manual.


lib/rubocop/cop/sorbet/forbid_include_const_literal.rb
```diff
@@ -62,25 +53,6 @@ def on_send(node)
...
-
-            # Apply fix
-            indent = ' ' * node.loc.column
-            fix = "#{include_name} = #{node.child_nodes.first.source}\n#{indent}"
-            corrector.insert_before(node.loc.expression, fix)
-            corrector.replace(node.child_nodes.first.loc.expression, include_name)
-          end
-        end
-
         # Find a free local variable name
```
paracycle, line 56: The `find_free_name` method can go as well.

paracycle (approved):
Technically, there is a better autocorrect for `ForbidIncludeConstLiteral` which changes `include Foo.helpers` to `T.unsafe(self).include Foo.helpers` but we don't have to add it. I guess it is better to warn about it and not modify the code.

This looks good! Thanks

PR #47, opened 2020-10-08
Fix type alias cop mismatching class/module/struct creation

paracycle:
Our `BindingConstantWithoutTypeAlias` has started mismatching classes/modules/structs defined via `Class.new`, `Module.new` and `Struct.new`, etc with a block that contains `T.xxx` inside. 

We need to explicitly match those dynamic module creation calls and exclude them.


Morriar (approved):

Morriar:
Good find!

paracycle:
@at Simplified the tests by triggering with `T.any` instead of `T.untyped`.

paracycle:
@Morriar Simplified the tests by triggering with `T.any` instead of `T.untyped`.

PR #50, opened 2021-01-11
Simplify `enforce_signatures`

### LINKED ISSUE #49: Inline visibility modifiers cause false positives in Sorbet/EnforceSignatures
To repro, create `a.rb` with:
```ruby
# typed: strict

class A
  sig { void }
  private def one
    1
  end
end
```
Expected result: no `Sorbet/EnforceSignatures` violation
Actual result:
```sh
$ rubocop --only Sorbet/EnforceSignatures a.rb
Inspecting 1 file
C

Offenses:

a.rb:5:11: C: [Correctable] Sorbet/EnforceSignatures: Each method is required to have a signature.
  private def one ...
          ^^^^^^^

1 file inspected, 1 offense detected, 1 offense auto-correctable
```

The autocorrect here is also a bit busted:
```
$ rubocop --only Sorbet/EnforceSignatures -a a.rb
Inspecting 1 file
C

Offenses:

a.rb:5:11: C: [Corrected] Sorbet/EnforceSignatures: Each method is required to have a signature.
  private def one ...
          ^^^^^^^
a.rb:6:11: C: [Corrected] Sorbet/EnforceSignatures: Each method is required to have a signature.
          def one ...
          ^^^^^^^

1 file inspected, 2 offenses detected, 2 offenses corrected
$ cat a.rb
# typed: strict

class A
  sig { void }
  private sig { returns(T.untyped) }
          sig { returns(T.untyped) }
          def one
    1
  end
end

### PR DESCRIPTION
**Pull request by Morriar**
Just look if there is a dangling signature in the current scope when we encounter a new def or attribute definition.

Fixes #49.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
+++ b/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
@@ -27,6 +27,11 @@ module RuboCop
       # * `ParameterTypePlaceholder`: placeholders used for parameter types (default: 'T.untyped')
       # * `ReturnTypePlaceholder`: placeholders used for return types (default: 'T.untyped')
       class EnforceSignatures < SignatureCop
+        def initialize(config = nil, options = nil)
+          super(config, options)
+          @last_sig_for_scope = {}
+        end
+
         def_node_matcher(:accessor?, <<-PATTERN)
           (send nil? {:attr_reader :attr_writer :attr_accessor} ...)
         PATTERN
@@ -40,8 +45,11 @@ module RuboCop
         end
 
         def on_send(node)
-          return unless accessor?(node)
-          check_node(node)
+          check_node(node) if accessor?(node)
+        end
+
+        def on_block(node)
+          @last_sig_for_scope[scope(node)] = node if signature?(node)
         end
 
         def autocorrect(node)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by KaanOzkan — APPROVED** [review]
LGTM pending the rubocop offense.

**Review by RyanBrushett — APPROVED** [review]

File: lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb [hunk 1]
```diff
@@ -63,22 +71,23 @@ def autocorrect(node)
           end
         end
 
+        def scope(node)
+          return nil unless node.parent
+          return node.parent if [:begin, :class, :module].include?(node.parent.type)
```
paracycle, line 76: What about blocks?

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
```diff
@@ -45,6 +45,48 @@ def foo; end
...
+          sig { void }
+          def foo; end
+        end
+
+        sig { void }
+        sig { void }
+        def foo; end
+      RUBY
+    end
+    it 'handles scopes correctly' do
```
paracycle, line 64: Should we test sigs in blocks as well?

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
Morriar: Good catch, I was storing the sig in the block's parent scope!

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
Morriar: Added a test

paracycle (approved):

PR #51, opened 2021-01-19
Handle index send events without changing global state

paracycle:
Thanks to [comments from @mbj](https://github.com/Shopify/rubocop-sorbet/pull/41#discussion_r558806672) I've realized that it is possible to reparse node source with index send emitting (and more modern features enabled) by using an anonymous subclass of the AST Builder class with modern features turned on, instead of playing around with the state of the global AST Builder object.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
@@ -55,7 +55,7 @@ module RuboCop
           return nil unless can_autocorrect?
 
           lambda do |corrector|
-            tree = call_chain(node_with_index_sends(node))
+            tree = call_chain(node_reparsed_with_modern_features(node))
               .sort_by { |call| ORDER[call.method_name] }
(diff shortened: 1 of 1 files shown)


cjhutchi (approved):

paracycle:
@mbj Thanks for your super detailed comment and your corrections on this PR. Also, apologies that I got sidetracked by other things and dropped this for so long.

I see the problems that could be happening due to partial reparsing of the node source, but, IMO, the context in which we do this operation is so narrow (inside of a `sig` block) that we don't expect any random Ruby code to exist there, let alone anything that might look like an lvar reference. Moreover, we haven't been "caught" by the previous, more hacky, version of this operation anyway, so at best this PR just improves things slightly.

Having said that, I will reach out to the `rubocop-ast` folks to see if it is possible to make them switch to the modern AST format or to allow cops to declare that they need certain features to work properly. If we can make progress there, I will revisit this issue.

Many thanks again for helping me make this better. Much appreciated! ❤️

Morriar (approved):
TIL

PR #54, opened 2021-03-06
Add cop for empty class/module definitions in RBI files

RyanBrushett:
Empty class/module definitions in RBI files should be expressed on a single line rather than multiple lines.

This is my first crack at writing one of these! Keen for any feedback.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -107,6 +107,13 @@ Sorbet/SignatureBuildOrder:
   Enabled: true
   VersionAdded: 0.3.0
 
+Sorbet/SingleLineRbiClassModuleDefinitions:
+  Description: 'Empty class and module definitions in RBI must be on a single line.'
+  Enabled: false
+  VersionAdded: '0.86'
+  Include:
+  - "**/*.rbi"
+
 Sorbet/StrictSigil:
   Description: 'All files must be at least at strictness `strict`.'
   Enabled: false
--- a/lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
+++ b/lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop ensures empty class/module definitions in RBI files are
+      # done on a single line rather than being split across multiple lines.
+      #
+      # @example
+      #
+      #   # bad
+      #   module SomeModule
+      #   end
+      #
+      #   # good
+      #   module SomeModule; end
+      class SingleLineRbiClassModuleDefinitions < RuboCop::Cop::Cop
+        MSG = 'Empty class/module definitions in RBI files should be on a single line.'
+
+        def on_module(node)
+          process_node(node)
+        end
+
+        def on_class(node)
+          process_node(node)
+        end
+
+        def autocorrect(node)
+          -> (corrector) { corrector.replace(node, convert_newlines(node.source)) }
+        end
+
+        protected
+
+        def convert_newlines(source)
+          source.sub(/[\r\n]+\s*[\r\n]*/, "; ")
+        end
+
+        def process_node(node)
+          return if node.body
+          return if node.single_line?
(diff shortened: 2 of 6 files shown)


config/default.yml
```diff
@@ -107,6 +107,13 @@ Sorbet/SignatureBuildOrder:
   Enabled: true
   VersionAdded: 0.3.0
 
+Sorbet/SingleLineRbiClassModuleDefinitions:
+  Description: 'Empty class and module definitions in RBI must be on a single line.'
+  Enabled: true
```
paracycle, line 112: Can we have this disabled by default? This is the style we prefer at Shopify but it might not be the style others would like, so I'd rather this be an opt-in for the library.

lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
```diff
@@ -0,0 +1,47 @@
...
+      #
+      # @example
+      #
+      #   # bad
+      #   module SomeModule
+      #   end
+      #
+      #   # good
+      #   module SomeModule; end
+
```
paracycle, line 17:
```suggestion
```

spec/rubocop/cop/sorbet/single_line_rbi_class_module_definitions_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+        class AnotherClass < SomeParentClass; end
+      RBI
+    end
+
+    it 'does not register an offence when a module is not empty' do
+      expect_no_offenses(<<~RBI)
+        module MyModule
+          def hello; end
+        end
+      RBI
```
paracycle, line 56: Can we also test with a multi-line method body:
```suggestion
        module MyModule
          def hello; end
        end

        module AnotherModule
          def world
          end
        end
      RBI
```

paracycle (approved):
Looks good, just a couple of minor comments.

spec/rubocop/cop/sorbet/single_line_rbi_class_module_definitions_spec.rb
RyanBrushett: Good call!

config/default.yml
RyanBrushett: Makes sense! Will do.

lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+
+        def autocorrect(node)
+          -> (corrector) { corrector.replace(node, convert_newlines(node.source)) }
+        end
+
+        protected
+
+        def convert_newlines(source)
+          source.sub("\n", "; ").tr("\n", "")
```
Morriar, line 35: What if the bock contains a space like this:

```rb
class Foo
  # <- empty indented line
end
```

Are we letting this being caught by another cop?

Morriar (approved):

lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
vinistock: Suggestion for handling the multiple empty lines:
```ruby
source.gsub(/[\r\n]+/, "; ")
```

vinistock (approved):

lib/rubocop/cop/sorbet/single_line_rbi_class_module_definitions.rb
RyanBrushett: Ah good eye! I'll see if I can make this a bit more clever.

RyanBrushett:
@vinistock @Morriar cleaned up the regex in line with your suggestions to make it a bit more clever and added another test.
Now it matches a one-or-more linebreaks, followed by none-or-more spaces and linebreaks.

One more quick 👀 at the latest commit please before I ⛵ ?

vinistock:
LGTM

Morriar (approved):

PR #55, opened 2021-03-09
Add one ancestor per requires_ancestor cop

RyanBrushett:
We want to be able to enforce that each requires_ancestor call is on a separate line, rather than chained together.

```ruby
# bad
module SomeModule
  requires_ancestor Kernel, Minitest::Assertions
end

# good
module SomeModule
  requires_ancestor Kernel
  requires_ancestor Minitest::Assertions
end
```

### DIFF
--- a/lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
+++ b/lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
@@ -1,0 +1,75 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+require 'rubocop'
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop ensures one ancestor per requires_ancestor line
+      # rather than chaining them as a comma-separated list.
+      #
+      # @example
+      #
+      #   # bad
+      #   module SomeModule
+      #     requires_ancestor Kernel, Minitest::Assertions
+      #   end
+      #
+      #   # good
+      #   module SomeModule
+      #     requires_ancestor Kernel
+      #     requires_ancestor Minitest::Assertions
+      #   end
+      class OneAncestorPerLine < RuboCop::Cop::Cop
+        MSG = 'Cannot require more than one ancestor per line'
+
+        def_node_search :requires_ancestors, <<~PATTERN
+          (send nil? :requires_ancestor ...)
+        PATTERN
+
+        def_node_matcher :more_than_one_ancestor, <<~PATTERN
+          (send nil? :requires_ancestor const const+)
+        PATTERN
+
+        def_node_search :abstract?, <<~PATTERN
+          (send nil? :abstract!)
+        PATTERN
+
+        def on_module(node)
+          return unless node.body
+          return unless requires_ancestors(node)
+          process_node(node)
+        end
+
+        def on_class(node)
+          return unless abstract?(node)
+          return unless requires_ancestors(node)
+          process_node(node)
+        end
+
+        def autocorrect(node)
+          -> (corrector) do
+            ra_call = node.parent
+            split_ra_calls = ra_call.source.gsub(/,\s+/, new_ra_line(ra_call.loc.column))
+            corrector.replace(ra_call, split_ra_calls)
+          end
+        end
+
+        private
+
+        def process_node(node)
+          requires_ancestors(node).each do |ra|
+            add_offense(ra.child_nodes[1]) if more_than_one_ancestor(ra)
+          end
+        end
+
+        def new_ra_line(indent_count)
+          indents = " " * indent_count
+          indented_ra_call = "#{indents}requires_ancestor "
+          "\n#{indented_ra_call}"
+        end
+      end
+    end
+  end
+end
--- a/manual/cops_sorbet.md
+++ b/manual/cops_sorbet.md
@@ -258,6 +258,30 @@ sig { params(b: String, a: Integer).void }
 def foo(b:, a: 1); end
 ```
 
+## Sorbet/OneAncestorPerLine
+
+Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+--- | --- | --- | --- | ---
+Enabled | Yes | Yes  | - | -
+
+This cop ensures one ancestor per requires_ancestor line
+rather than chaining them as a comma-separated list.
+
+### Examples
+
+```ruby
+# bad
+module SomeModule
+  requires_ancestor Kernel, Minitest::Assertions
+end
+
+# good
+module SomeModule
+  requires_ancestor Kernel
+  requires_ancestor Minitest::Assertions
+end
+```
+
 ## Sorbet/ParametersOrderingInSignature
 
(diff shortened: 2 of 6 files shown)


manual/cops_sorbet.md
```diff
@@ -302,7 +326,7 @@ You can subclass it to use the `on_signature` trigger and the `signature?` node
 
 Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 --- | --- | --- | --- | ---
-Enabled | Yes | Yes  | 0.86 | -
+Disabled | Yes | Yes  | 0.86 | -
```
RyanBrushett, line 329: Forgot to regenerate the docs before shipping my previous PR.

manual/cops_sorbet.md
```diff
@@ -315,7 +339,7 @@ module SomeModule
 end
 
 # good
-module SomeModule;end
+module SomeModule; end
```
RyanBrushett, line 342: Forgot to regenerate the docs before shipping my previous PR.

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+        def on_class(node)
+          return unless abstract?(node)
+          return unless requires_ancestors(node)
+          process_node(node)
+        end
+
+        def autocorrect(node)
+          -> (corrector) do
+            ra_call = node.parent
+            split_ra_calls = ra_call.source.gsub(/,\s+/, new_ra_line(ra_call))
```
RyanBrushett, line 54: This felt simple enough to not justify putting it in its own class but I'm keen to hear what you all think.

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+      #     requires_ancestor Kernel, Minitest::Assertions
+      #   end
+      #
+      #   # good
+      #   module SomeModule
+      #     requires_ancestor Kernel
+      #     requires_ancestor Minitest::Assertions
+      #   end
+      class OneAncestorPerLine < RuboCop::Cop::Cop
+        MSG = 'Cannot declare more than one ancestor per line'
```
Morriar, line 25:
```suggestion
        MSG = 'Cannot require more than one ancestor per line'
```

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        def_node_search :requires_ancestors, <<~PATTERN
+          (send nil? :requires_ancestor ...)
```
Morriar, line 28: I think we only one to correct them if called on `self` (so always no receiver):

```suggestion
          (send nil :requires_ancestor ...)
```

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+        PATTERN
+
+        def on_module(node)
+          return unless node.body
+          return unless requires_ancestors(node)
+          process_node(node)
+        end
+
+        def on_class(node)
+          return unless abstract?(node)
```
Morriar, line 46: Do we really need this?

RyanBrushett: I guess we don't _need_ it but since requires_ancestor only makes sense from inside modules or abstract classes, I figured a quick return unless it's `abstract!` would be fine. Can certainly take it out!

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
RyanBrushett: Makes sense! I'll fix these in the other places too.

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        def_node_search :requires_ancestors, <<~PATTERN
+          (send nil? :requires_ancestor ...)
+        PATTERN
+
+        def_node_matcher :more_than_one_ancestor, <<~PATTERN
+          (send nil? :requires_ancestor const const+)
+        PATTERN
+
+        def_node_search :abstract?, <<~PATTERN
```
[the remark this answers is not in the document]
RyanBrushett: (responded below)

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
RyanBrushett: Hmm if I change these, all of my tests fail.
Will dig into it a bit.

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
```diff
@@ -0,0 +1,81 @@
...
+        def indents(count)
+          " " * count
+        end
+
+        def new_ra_line(ra_call)
+          out = StringIO.new
+          out << "\n"
+          out << indents(ra_call.loc.column) # Match indent
+          out << "requires_ancestor "
+          out.string
```
vinistock, line 76: NIT: this is a good opportunity to use `rjust(count, char)` (right justify). It justifies from the right using `count` number of `char`s. The default is blank spaces.

We can probably skip using `StringIO` too and use an unfrozen string.
```rb
out = +"\n" # create unfrozen string
out << "requires_ancestor ".rjust(ra_call.loc.column)
out
```

vinistock (approved):

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
RyanBrushett: Neat! TIL
Yeah I 'borrowed' the StringIO bit from another cop to make the `indents(...)` call a little 'cleaner' looking. `rjust` should help me clean this up! Thanks!

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
Morriar: My  bad, `nil` is actually the `nil` value... 🤦

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
RyanBrushett: Interesting! Rjust doesn't do what you'd expect.

```
rjust(integer, padstr=' ') → new_str

If integer is greater than the length of str, returns a new String of length integer with str right justified and padded with padstr; otherwise, returns str.

"hello".rjust(4) #=> "hello"
```

I will tidy this up a bit though.

lib/rubocop/cop/sorbet/one_ancestor_per_line.rb
Morriar: If it works I'm happy with the early return. Thanks for adding the test 👍

Morriar (approved):

PR #56, opened 2021-03-10
Add cop to forbid extend T::Sig and T::Helpers in shims

RyanBrushett:
It's unnecessary to `extend T::Sig` in shims. This cop allows us to prevent that and autocorrect it 👍

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -55,7 +55,7 @@ Sorbet/FalseSigil:
 
 Sorbet/ForbidExtendTSigInShims:
   Description: 'Forbid the use of `extend T::Sig` in RBI shims'
-  Enabled: false
+  Enabled: true
   VersionAdded: '0.86'
   Include:
   - "**/*.rbi"
--- a/config/default.yml
+++ b/config/default.yml
@@ -53,6 +53,13 @@ Sorbet/FalseSigil:
   - db/**/*.rb
   - script/**/*
 
+Sorbet/ForbidExtendTSigInShims:
+  Description: 'Forbid the use of `extend T::Sig` in RBI shims'
+  Enabled: false
+  VersionAdded: '0.86'
+  Include:
+  - "sorbet/rbi/shims/**/*.rbi"
+
 Sorbet/ForbidIncludeConstLiteral:
   Description: 'Forbids include of non-literal constants.'
   Enabled: false
--- a/config/default.yml
+++ b/config/default.yml
@@ -58,7 +58,7 @@ Sorbet/ForbidExtendTSigInShims:
   Enabled: false
   VersionAdded: '0.86'
   Include:
-  - "sorbet/rbi/shims/**/*.rbi"
+  - "**/*.rbi"
 
 Sorbet/ForbidIncludeConstLiteral:
   Description: 'Forbids include of non-literal constants.'
--- a/config/default.yml
+++ b/config/default.yml
@@ -53,8 +53,8 @@ Sorbet/FalseSigil:
   - db/**/*.rb
   - script/**/*
 
-Sorbet/ForbidExtendTSigInShims:
-  Description: 'Forbid the use of `extend T::Sig` in RBI shims'
+Sorbet/ForbidExtendTSigHelpersInShims:
+  Description: 'Forbid the use of `extend T::Sig` and `extend T::Helpers` in RBI shims'
   Enabled: true
   VersionAdded: '0.86'
   Include:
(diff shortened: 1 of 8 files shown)


config/default.yml
```diff
@@ -53,6 +53,13 @@ Sorbet/FalseSigil:
   - db/**/*.rb
   - script/**/*
 
+Sorbet/ForbidExtendTSigInShims:
+  Description: 'Forbid the use of `extend T::Sig` in RBI shims'
+  Enabled: false
+  VersionAdded: '0.86'
+  Include:
+  - "sorbet/rbi/shims/**/*.rbi"
```
Morriar, line 61: I think it's safe to include all kind of RBI by default. Even generator should not do this.

```suggestion
  - "**/*.rbi"
```

Morriar (approved):

config/default.yml
```diff
@@ -53,6 +53,13 @@ Sorbet/FalseSigil:
   - db/**/*.rb
   - script/**/*
 
+Sorbet/ForbidExtendTSigInShims:
+  Description: 'Forbid the use of `extend T::Sig` in RBI shims'
+  Enabled: false
```
paracycle, line 58: We can safely enable this as well. If anyone is doing it, they are doing it wrong.

manual/cops_sorbet.md
```diff
@@ -315,7 +347,7 @@ module SomeModule
 end
 
 # good
-module SomeModule;end
+module SomeModule; end
 ```
```
paracycle, line 351: Hmm, were we not keeping the docs in sync with the source? I guess this was not properly updated in the last PR, correct?

spec/rubocop/cop/sorbet/forbid_extend_t_sig_in_shims_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        end
+      RBI
+    end
+  end
+
+  describe('no offences') do
+    it 'does not add an offence to uses of extend that are not T::Sig' do
+      expect_no_offenses(<<~RBI)
+        module MyModule
+          extend ActiverSupport::Concern
```
paracycle, line 26:
```suggestion
          extend ActiveSupport::Concern
```

`Activer` is good but we want just `Active` 😂

spec/rubocop/cop/sorbet/forbid_extend_t_sig_in_shims_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+          def foo; end
+        end
+      RBI
+    end
+  end
+
+  describe('autocorrect') do
+    it 'autocorrects usages of extend T::Sig by removing them' do
```
paracycle, line 35: Can we add a case where the body of the module/class just includes a single `extend T::Sig`?

vinistock (approved):

manual/cops_sorbet.md
RyanBrushett: My other PR that I merged should have this updated too tbh.

RyanBrushett:
We talked IRL, I'll expand this to catch T::Helpers and T::Generics as well.

spec/rubocop/cop/sorbet/forbid_extend_t_sig_in_shims_spec.rb
Morriar: Here rooting for `ActivestSupport` 💪

spec/rubocop/cop/sorbet/forbid_extend_t_sig_in_shims_spec.rb
Morriar: And maybe `extend(T::Sig)`

RyanBrushett:
Forgot to rename things before I requested a review 🤦 doing that now.

Morriar (approved):

lib/rubocop/cop/sorbet/forbid_extend_t_sig_helpers_in_shims.rb
```diff
@@ -0,0 +1,53 @@
...
+      #
+      #   # good
+      #   module SomeModule
+      #     sig { returns(String) }
+      #     def foo; end
+      #   end
+      class ForbidExtendTSigHelpersInShims < RuboCop::Cop::Cop
+        include RangeHelp
+
+        MSG = 'Extending T::Sig or T::Helpers in a shim is unnecessary'
```
vinistock, line 28: Question: are there other things Sorbet allows you to extend that we want to avoid too? Like `T::Generic` maybe?

RyanBrushett: We talked about T::Generic actually telling the user a lot about the class the shim is representing so we figured we'd let that fly. But T::Sig and T::Helpers don't really add anything to a shim.

PR #59, opened 2021-04-28
Add scope helper position cop

vinistock:
Add the new cop to enforce that no method definitions appear before invoking `interface!`, `abstract!` or `final!`.

There are two things in particular I'd appreciate feedback.
1. Is there a cleaner way of removing surrounding blank lines from the matched node? Basically, if there is an empty line before and after the scope helper, I want to remove both and still leave a space between method definitions. I was able to get that functionality, but it's a bit messy
2. Is the way I matched the send_node to `abstract!`, `interface!` or `final!` okay or is there a nicer way of defining that?

### DIFF
--- a/lib/rubocop/cop/sorbet/scope_helper_position.rb
+++ b/lib/rubocop/cop/sorbet/scope_helper_position.rb
@@ -31,7 +31,7 @@ module RuboCop
       #   end
       class ScopeHelperPosition < RuboCop::Cop::Cop
         include RangeHelp
-        HELPERS = %i(interface! abstract! final!).freeze
+        HELPERS = %i(interface! abstract! final! sealed!).freeze
 
         def autocorrect(node)
           lambda do |corrector|
--- a/lib/rubocop/cop/sorbet/scope_helper_position.rb
+++ b/lib/rubocop/cop/sorbet/scope_helper_position.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop ensures that helpers such as abstract!, interface! and final!
+      # are invoke before any method definitions
+      #
+      # @example
+      #
+      #   # bad
+      #   module Interface
+      #     extend T::Sig
+      #     extend T::Helpers
+      #
+      #     sig { returns(String) }
+      #     def foo; end
+      #
+      #     interface!
+      #   end
+      #
+      #   # good
+      #   module Interface
+      #     extend T::Sig
+      #     extend T::Helpers
+      #
+      #     interface!
+      #
+      #     sig { returns(String) }
+      #     def foo; end
+      #   end
+      class ScopeHelperPosition < RuboCop::Cop::Cop
+        include RangeHelp
+        HELPERS = %i(interface! abstract! final!).freeze
+
+        def autocorrect(node)
+          lambda do |corrector|
+            end_pos = if processed_source[node.source_range.line].blank? &&
+              processed_source[node.source_range.line - 2].blank?
+              node.source_range.end_pos + 1
+            else
+              node.source_range.end_pos
+            end
+
+            range = range_between(node.source_range.begin_pos, end_pos)
+            corrector.remove(range_by_whole_lines(range, include_final_newline: true))
+
+            first_def = node.parent.each_child_node.find(&:def_type?)
+            first_sig = node.parent.each_child_node.find { |child| child.method_name == :sig }
+            before_node = first_sig && first_sig.first_line < first_def.first_line ? first_sig : first_def
+            indentation = " " * before_node.loc.column
+
+            corrector.insert_before(before_node, "#{node.source}\n\n#{indentation}")
+          end
+        end
+
+        def on_send(node)
+          return unless HELPERS.include?(node.method_name)
+
+          node.parent.each_child_node do |child_node|
+            if child_node.def_type? && child_node.first_line < node.first_line
+              add_offense(node, message: "Cannot invoke #{node.method_name} after method definitions")
+              break
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/sorbet/scope_helper_position.rb
+++ b/lib/rubocop/cop/sorbet/scope_helper_position.rb
@@ -32,7 +32,7 @@ module RuboCop
       class ScopeHelperPosition < RuboCop::Cop::Cop
         include RangeHelp
         HELPERS = %i(interface! abstract! final! sealed!).freeze
-        ALLOWED_METHODS = %i(extend include).freeze
+        ALLOWED_METHODS = %i(extend include prepend).freeze
 
         def autocorrect(node)
           lambda do |corrector|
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/sorbet/scope_helper_position.rb
```diff
@@ -0,0 +1,70 @@
...
+      #     extend T::Helpers
+      #
+      #     interface!
+      #
+      #     sig { returns(String) }
+      #     def foo; end
+      #   end
+      class ScopeHelperPosition < RuboCop::Cop::Cop
+        include RangeHelp
+        HELPERS = %i(interface! abstract! final!).freeze
```
Morriar, line 34: `sealed!` should also be here

lib/rubocop/cop/sorbet/scope_helper_position.rb
```diff
@@ -0,0 +1,70 @@
...
+              processed_source[node.source_range.line - 2].blank?
+              node.source_range.end_pos + 1
+            else
+              node.source_range.end_pos
+            end
+
+            range = range_between(node.source_range.begin_pos, end_pos)
+            corrector.remove(range_by_whole_lines(range, include_final_newline: true))
+
+            first_def = node.parent.each_child_node.find(&:def_type?)
```
Morriar, line 48: It might be more robust to ensure only `extend`, `prepend` and `include` are before?

Without this you'll also have to deal with calls to singleton methods, blocks, sub-scopes, ...

paracycle: I think @Morriar means "calls to" those methods will still be above where you will ultimately insert the helper method call, as in:
```ruby
class Foo
  extend T::Helpers

  attribute :foo, String
  attribute :bar, Integer

  validate_uniqueness :foo, if: -> do
    # some condition here
  end

  some_method_call_that_takes_a_block do 
    # some long lines of code here
  end

  interface!

  def foo_bar
  end
end
```
the `interface!` call is still too low in the class definition even though it is above the first method call.

Can we ensure that the helper method call is right below the `extend T::Helpers` line and that that itself is right below `extend T::Sig`?

```ruby
class Foo
  extend T::Sig

  some_class_method_call_here

  sig { void }
  def foo_bar

  extend T::Helpers
 
  sealed!
end
```
should be turned into:  
```ruby
class Foo
  extend T::Sig
  extend T::Helpers
 
  sealed!

  some_class_method_call_here

  sig { void }
  def foo_bar

end
```

lib/rubocop/cop/sorbet/scope_helper_position.rb
```diff
@@ -0,0 +1,71 @@
...
+      #
+      #     interface!
+      #
+      #     sig { returns(String) }
+      #     def foo; end
+      #   end
+      class ScopeHelperPosition < RuboCop::Cop::Cop
+        include RangeHelp
+        HELPERS = %i(interface! abstract! final! sealed!).freeze
+        ALLOWED_METHODS = %i(extend include).freeze
```
Morriar, line 35: And `prepend` I guess

lib/rubocop/cop/sorbet/scope_helper_position.rb
```diff
@@ -0,0 +1,71 @@
...
+
+            if processed_source[node.source_range.line - 2].empty? &&
+              (processed_source[node.source_range.line].empty? || processed_source[node.source_range.line] == "end")
+              corrector.remove(
+                range_by_whole_lines(source_range(processed_source.buffer, node.source_range.line - 1, 0))
+              )
+            end
+
+            before_node = node.parent.each_child_node.find do |child_node|
+              !ALLOWED_METHODS.include?(child_node.method_name)
```
Morriar, line 49: Would that mean I could write something like this?

```rb
class Foo
  def include; end
  abstract!
end
```

spec/rubocop/cop/sorbet/scope_helper_position_spec.rb
```diff
@@ -0,0 +1,237 @@
...
+          def foo; end
+
+          def bar; end
+        end
+      CORRECTED
+
+      expect(autocorrect_source(source)).to(eq(corrected_source))
+    end
+
+    it("autocorrects when there are method invocations") do
```
Morriar, line 211: Can we add a test with all the helpers in the same scope to ensure we don't create an infinite loop in a later change? 🙏 

```rb
class Foo
  extend T::Helpers

  def foo; end

  interface!
  abstract!
  sealed!
  final!
end
```

Morriar: As for maximum coverage we might also want a test when there is no `include/extend/prepend`:

```rb
class Foo
  def  foo; end # with or without this

  interface! # and maybe 
  abstract! # a few inlined
  sealed! # comments
  final! # to ensure the right behaviour
end
```

vinistock:
I fixed a few issues that I found with the implementation of the cop. Namely,
1. allowed requires_ancestor to be invoked before scope helpers
2. fixed broken behavior when using multiple scope helpers
3. fixed broken behavior for block invocations

One caveat that I noticed and would appreciate feedback on, is that if methods are invoked before allowed methods, then we still move the scope helper. I believe this is the correct behavior and we shouldn't try to fix the order of the rest of the file, but I'd love to hear opinions. This is what I mean:
```ruby
class Post
  # this is a method called table_name=, which means we will move the scope helper above it
  self.table_name = "articles"

  # because there's a method invocation above this include, the scope helper will actually end up above it
  include Something

  abstract!
end
```

The reason I want feedback is because I believe this scenario below will be problematic:
```ruby
# Because of the method invocation, this class would be corrected
class Post
  self.table_name = "articles"
  extend T::Helpers
  abstract!
end

# The corrected file is actually incorrect, since we cannot invoke the scope helper before including T::Helpers
class Post
  abstract!
  self.table_name = "articles"
  extend T::Helpers
end
```
I'm not sure how we can handle this honestly, since it's hard to know if other special methods like `table_name` will be invoked before includes and extends. Ideas?

Morriar:
What if we ensure that `abstract` and friends always come right after the `extend T::Helpers` would that make more sense for code like your example?

vinistock:
But what happens if someone does this? The only way to fix it is if we also move the `extend` properly, which I believe is a bit much. Feels like trying to fix too many things in the same cop.
```ruby
class Post
  extend T::Sig

  sig { abstract.void }
  def abstract; end

  extend T::Helpers
  
  sig { abstract.void }
  def another_abstract; end

  interface!
end
```

vinistock:
The latest commit introduces functionality to figure out if there's an `extend T::Helpers` after the node we are going to move the scope helper to and then moves the `extend` statement along. Let me know your thoughts.

Morriar:
As discussed IRL, we'll move the `extend T::Helpers` on top of the scope then move the helpers below it.

lib/rubocop/cop/sorbet/scope_helper_position.rb
```diff
@@ -0,0 +1,89 @@
...
+              )
+            end
+
+            before_node = node.parent.each_child_node.find do |child_node|
+              method_invocation = child_node.send_type? || child_node.block_type?
+              offense_invocation = method_invocation && !ALLOWED_METHODS.include?(child_node.method_name) &&
+                !HELPERS.include?(child_node.method_name)
+
+              child_node.def_type? || offense_invocation
+            end
```
paracycle, line 54: This seems to be common logic between the autocorrect and the check. Might be a good idea to extract this into a method.

vinistock:
Closing this PR for now. We might take a different approach that may not need these new cops.

PR #60, opened 2021-04-30
Add callback conditionals cop

vinistock:
This PR does two things in different commits

1. Add `byebug` to the project. It's hard to figure out small details without it 😅 
2. Add the callback conditionals binding cop, to ensure that blocks are type checked against the right types

E.g.:
```ruby
# Bad. Sorbet is not able to determine that the block will be executed in the context of an instance
# and then thinks `should?` is undefined
class Post < ApplicationRecord
  before_create :do_it, if: -> { should? }
end

# Good. By binding it to the right type, Sorbet is able to type check the block
class Post < ApplicationRecord
  before_create :do_it, if: -> { T.bind(self, Post).should? }
end
```

### DIFF
--- a/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
+++ b/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
@@ -26,8 +26,46 @@ module RuboCop
       #     end
       #   end
       class CallbackConditionalsBinding < RuboCop::Cop::Cop
-        CALLBACKS = %i(before_create).freeze
-        include RangeHelp
+        CALLBACKS = %i(
+          before_create
+          before_save
+          before_destroy
+          before_update
+
+          after_create
+          after_save
+          after_destroy
+          after_update
+          after_touch
+          after_initialize
+          after_find
+
+          around_create
+          around_save
+          around_destroy
+          around_update
+
+          before_commit
+
+          after_commit
+          after_create_commit
+          after_destroy_commit
+          after_rollback
+          after_save_commit
+          after_update_commit
+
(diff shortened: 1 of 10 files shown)


lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
```diff
@@ -0,0 +1,90 @@
...
+      #   # good
+      #   class Post < ApplicationRecord
+      #     before_create :do_it, if: -> { T.bind(self, Post).should_do_it? }
+      #
+      #     def should_do_it?
+      #       true
+      #     end
+      #   end
+      class CallbackConditionalsBinding < RuboCop::Cop::Cop
+        CALLBACKS = %i(before_create).freeze
```
Morriar, line 29: I'm surprised it's the only callback?

Morriar (approved):

lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
```diff
@@ -26,8 +26,46 @@ module Sorbet
       #     end
       #   end
       class CallbackConditionalsBinding < RuboCop::Cop::Cop
-        CALLBACKS = %i(before_create).freeze
-        include RangeHelp
+        CALLBACKS = %i(
+          before_create
```
Morriar, line 30: Seems like we were missing quite a few of them actually 😂

Morriar (approved):
Nice, thanks! ❤️

RyanBrushett (approved):
Looks good!

PR #62, opened 2021-05-05
Add extend sig position cop

vinistock:
This PR is the first of 3 cops related to #59

We realized that, in order to be able to safely autocorrect scope helpers, we need to make sure that the first line in classes/modules is `extend T::Sig`, the second `extend T::Helpers` and then the scope helper. Other attempts to autocorrect could lead to broken code after the correction.

Splitting this into 3 cops works better, since each cop is only responsible to autocorrect a single type of node.

This PR adds the cop to correct the position of `extend T::Sig` and always move it to the first thing inside a class or module.

The next PR will include the `extend T::Helpers` cop and finally we'll adapt #59 to only take care of the scope helpers.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -43,6 +43,11 @@ Sorbet/EnforceSignatures:
   Enabled: false
   VersionAdded: 0.3.4
 
+Sorbet/ExtendSigPosition:
+  Description: 'Ensures extend T::Sig is always invoked first in classes/modules.'
+  Enabled: true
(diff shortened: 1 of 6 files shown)


Morriar (approved):
Where will be placed the `extend T::Helpers` related to `extend T::Sig`?

vinistock:
Closing this PR for now. We might take a different approach that may not need these new cops.

PR #66, opened 2021-05-28
Add ForbidTUnsafe cop

kddeisz:
I don't think this cop should be enabled by default, as it's a very common and convenient way to bail out of type checks when Sorbet can't figure out what's going on. However, there are times when you want to find all usage of `T.unsafe` in your codebase. For example, when tapioca releases new DSLs that make it so that more things can be typed, a lot of the time that means we can get rid of things that we previously called `T.unsafe`. So this cop is mostly to find and document those usages so that they don't get lost.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -86,6 +86,12 @@ Sorbet/ForbidSuperclassConstLiteral:
   Exclude:
   - db/migrate/*.rb
 
+Sorbet/ForbidTUnsafe:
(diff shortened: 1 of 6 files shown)


Morriar (approved):

Morriar (approved):

Morriar:
Thanks @kddeisz! ❤️

PR #70, opened 2021-06-05
Add version bounds to unparser dependency

mbj:
Unparser `0.5` is not compatible with recent ruby releases. And to avoid frustration we should pin to the latest stable version.

### DIFF
--- a/rubocop-sorbet.gemspec
+++ b/rubocop-sorbet.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_development_dependency("rspec", "~> 3.7")
-  spec.add_development_dependency("unparser")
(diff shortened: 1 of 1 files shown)


mbj:
I've actually signed the CLA, but the status did not update here.

mbj:
I'd also like to add that `0.6` series was greatly reduced external dependencies. So nailing it makes lots of sense to avoid dep issues.

mbj:
@Morriar I do and will have no time to make a passing PR. I mostly figured I open a PR over an issue. Chances you can take it from here? 

I had users of `unparser` that use it via `rubocop-sorbet` contact me for issues, and the fix could be `rubocop-sorbet` specifying a version as these issues than at least belong to a maintained version of unparser (I abandoned below 0.6 for patches).

mbj:
@Morriar Thanks for removing this from my brain!

PR #74, opened 2021-06-30
Always correct conditional to multiline in CallbackConditionalsBinding

vinistock:
This PR applies a few fixes to the `CallbackConditionalsBinding` cop.

**Always breaks lines for multi statement procs**
```ruby
# this
if: -> { something && other_thing }

# no longer is corrected to
if: -> { T.bind(self, Class); something && other_thing }

# And is instead broken down to
if: -> {
  T.bind(self, Class)
  something && other_thing
}
```

**Substitute T.unsafe usage for bind**
```ruby
# this
if: -> { T.unsafe(self).something? }

# becomes
if: -> { T.bind(self, Class).something? }
```

**Array conditionals are not considered offenses anymore**
```ruby
# this is okay
if: [:method1, :method2]
```

**Procs that accept arguments are not considered offenses**
```ruby
# this could only be corrected when the receiver is the proc's argument, but it gets pretty complicated
# Since these don't produce type errors, I think it's okay to not try to auto correct it
if: ->(record) { invoke_method(record) }
```

**Conditionals where the recursive receiver is not self are now corrected to multiline**
```ruby
# this
if: -> { %w(a b).include?(some_method) }

# becomes
if: -> {
  T.bind(self, Class)
  %w(a b).include?(some_method)
}
```

**Chained send statements are properly corrected in a single line**
```ruby
# this
if: -> { something.present? }

# becomes
if: -> { T.bind(self, Class).something.present? }
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -63,4 +63,4 @@ DEPENDENCIES
   yard (~> 0.9)
 
 BUNDLED WITH
-   2.2.16
+   2.2.21
--- a/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
+++ b/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
@@ -52,13 +52,38 @@ module RuboCop
             _, _, block = conditional.child_nodes
             expected_class = node.parent_module_name
 
+            if block.source.include?("T.unsafe(self).")
+              corrector.replace(block, block.source.gsub("T.unsafe(self)", "T.bind(self, #{expected_class})"))
+              return
+            end
+
             bind = if block.begin_type?
               indentation = " " * block.child_nodes.first.loc.column
               "T.bind(self, #{expected_class})\n#{indentation}"
-            elsif block.child_nodes.empty? && !block.ivar_type?
+            elsif block.send_type? && recursive_receiver_is_self?(block)
               "T.bind(self, #{expected_class})."
             else
-              "T.bind(self, #{expected_class}); "
+              do_end_lambda = conditional.source.include?("do") && conditional.source.include?("end")
+
+              unless do_end_lambda
+                # We are converting a one line lambda into a multiline
+                # Remove the space after the `{`
+                if /{\s/.match?(conditional.source)
+                  corrector.remove_preceding(block, 1)
+                end
+
+                # Remove the last space and `}` and re-add it with a line break
+                # and the correct indentation
+                base_indentation = " " * node.loc.column
+                chars_to_remove = /\s}/.match?(conditional.source) ? 2 : 1
+                corrector.remove_trailing(conditional, chars_to_remove)
+                corrector.insert_after(block, "\n#{base_indentation}}")
+              end
+
+              # Add the T.bind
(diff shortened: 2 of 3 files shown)


RyanBrushett (approved):
I dig it

lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
```diff
@@ -52,13 +52,38 @@ def autocorrect(node)
             _, _, block = conditional.child_nodes
             expected_class = node.parent_module_name
 
+            if block.source.include?("T.unsafe(self).")
```
Morriar, line 55: Can't it be as problem with this?

```rb
something: -> { T.unsafe(self).passing_splat(*args) }
```

Morriar: If we're going to change working code into non-working code yes.

Another solution is to move the `T.unsafe` part in another cop that is optional and just used to auto-correct the `T.unsafe` in `T.bind`

Gemfile.lock
```diff
@@ -63,4 +63,4 @@ DEPENDENCIES
   yard (~> 0.9)
 
 BUNDLED WITH
-   2.2.16
+   2.2.21
```
Morriar, line 66: Is this change expected?

vinistock: I assume it happened because my bundler version is newer. Would you prefer to roll this back? IMO, I'd just keep it

vinistock:
Added another test case for when two classes are defined inside the same namespace.
```ruby
module Something
  class Namespace::Klass
    # Properly corrects to Namespace::Klass because it's defined on the same line
    validates :it, if: -> { T.bind(self, Namespace::Klass).should? }
  end

  class AnotherOne
    # Properly corrects to AnotherOne because using the fully qualified name here is not necessary
    validates :it, if: -> { T.bind(self, AnotherOne).should? }
  end
end
```

Gemfile.lock
Morriar: Let's remove it 🔥

lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
```diff
@@ -50,15 +50,46 @@ def autocorrect(node)
...
+              do_end_lambda = conditional.source.include?("do") && conditional.source.include?("end")
+
+              unless do_end_lambda
+                # We are converting a one line lambda into a multiline
+                # Remove the space after the `{`
+                if /{\s/.match?(conditional.source)
+                  corrector.remove_preceding(block, 1)
+                end
+
+                # Remove the last space and `}` and re-add it with a line break
```
Morriar, line 81: Shouldn't the block be changed in a `do` .. `end` if it's going to be multiline? In our style guide we enforce [BlockDelimiters](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/BlockDelimiters).

Morriar: Fair 👍

Morriar (approved):

PR #76, opened 2021-09-24
Add cop rules for RBI checking

Morriar:
A collection of Rubocop rules to be used explicitly to lint RBI files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,15 @@ RuboCop::RakeTask.new do |task|
 end
 ```
 
+### Rubocop rules for RBI files
+
+Put this into your `.rubocop.yml`:
+
+```yaml
+inherit_gem:
+  rubocop-sorbet: config/rbi.yml
+```
+
 ## The Cops
 All cops are located under [`lib/rubocop/cop/sorbet`](lib/rubocop/cop/sorbet), and contain examples/documentation.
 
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):

README.md
```diff
@@ -60,6 +60,19 @@ RuboCop::RakeTask.new do |task|
...
+
+Put this into your `.rubocop.yml`:
+
+```yaml
+inherit_gem:
+  rubocop-sorbet: config/rbi.yml
+
+AllCops:
+  Include:
+    - '**/*.rbi'
```
paracycle, line 73: Can't we have this in `rbi.yml` so that it only applies to `**/*.rbi` files by default?

config/rbi.yml
```diff
@@ -0,0 +1,254 @@
...
+  Enabled: true
+
+Sorbet/SingleLineRbiClassModuleDefinitions:
+  Enabled: true
+
+Sorbet/ValidSigil:
+  Enabled: true
+  RequireSigilOnAllFiles: true
+  MinimumStrictness: false
+  SuggestedStrictness: true
```
paracycle, line 216: Are these different from the default rules that we already export from this gem?

Morriar: Yes, some of them are disabled by default in the original config (`ForbidIncludeConstLiteral`, `ForbidSuperConstLiteral`, `TrueSigil`...). Others are enabled in the default config but we don't want them here or we want them with a different config.

PR #80, opened 2021-10-20
Add a cop to enforce a single Sorbet sigil per file

RyanBrushett:
Resolves https://github.com/Shopify/sorbet-issues/issues/462

I made this autocorrectable with the assumption that [the first sigil is the `correct` one](https://sorbet.run/#%23%20typed%3A%20true%0A%23%20typed%3A%20false%0A%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28x%3A%20Integer%29.void%7D%0Adef%20foo%28x%29%0A%20%20puts%28x%20%2B%201%29%0Aend%0A%0Affoo%280%29%0Afoo%28%22not%20an%20int%22%29) but we could scrap the autocorrect entirely, forcing the user to investigate. Thoughts?

Also: I extended `ValidSigil` since that seemed like it made sense and aligns with how `EnforceSigilOrder` is built. Thoughts?

See tests!


config/default.yml
```diff
@@ -38,6 +38,11 @@ Sorbet/EnforceSigilOrder:
   Enabled: true
   VersionAdded: 0.3.4
 
+Sorbet/sigils/EnforceSingleSigil:
+  Description: 'Ensures that there is only one Sorbet sigil in a file.'
+  Enabled: true
+  VersionAdded: '<<next>>'
```
vinistock, line 44: TIL. What does this do?

vinistock (approved):

config/default.yml
RyanBrushett: Haha no, I wish. That'd be real neat.

PR #81, opened 2021-10-21
Mark callback binding cop as unsafe and disable it by default

vinistock:
The callback binding cop has an unsafe auto-correct, since some libraries define similar style callbacks using the same names as Rails, but that need different types of bindings. This can lead to false positives and introduce typing errors when running auto-correct.

This PR just disables the cop by default, marks it as unsafe and adds a note about why in the documentation.

### DIFF
--- a/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
+++ b/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
@@ -7,7 +7,7 @@ module RuboCop
       # so that they are type checked properly.
       #
       # Auto-correction is unsafe because other libraries define similar style callbacks as Rails, but don't always need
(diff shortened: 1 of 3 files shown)


lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
```diff
@@ -6,6 +6,10 @@ module Sorbet
       # This cop ensures that callback conditionals are bound to the right type
       # so that they are type checked properly.
       #
+      # Auto-correction is unsafe because other libraries define similar style callbacks as Rails, but don't always need
+      # binding to the attached class. Auto-correcting those usages can lead to false positives and auto-corretion
```
paracycle, line 10:
```suggestion
      # binding to the attached class. Auto-correcting those usages can lead to false positives and auto-correction
```

Issue #79, opened 2021-10-19
ForbidRBIOutsideOfSorbetDir conflicts with sorbet docs, lacks guidance


dduugg:
The sorbet [docs](https://sorbet.org/docs/rbi#rbis-within-gems) advise gems to "include their external interfaces RBI files in an `rbi/` directory". This leads to `ForbidRBIOutsideOfSorbetDir` violations. (The cop would also benefit from providing guidance on structuring `sorbet/` with custom `rbi` files.)

dduugg:
Hi @Morriar thanks for the response. Based on my reading of the sorbet docs, I think that the error message would still be off with that configuration:

> RBI files are only accepted in the sorbet/rbi/ directory.

It also doesn't provide useful guidance as to how one should structure `sorbet/rbi/` with custom rbi files (I think this is the more salient issue, FWIW).

RyanBrushett:
With the addition of an `AllowedPaths` option for (what was) the ForbidRBIOutsideOfSorbetDir cop, the user can configure where they expect their RBI files to live, defaulting to `sorbet/rbi/**`. See #83 and [check out the doc entry for the `ForbidRBIOutsideOfAllowedPaths` cop here](https://github.com/Shopify/rubocop-sorbet/blob/master/manual/cops_sorbet.md#sorbetforbidrbioutsideofallowedpaths).

In terms of useful guidance: Opinions on structure should come from the Sorbet documentation or the tool used to generate RBI, (e.g. [Tapioca](https://github.com/shopify/tapioca)). It's up to the cop to be flexible enough to work with whatever you've settled on.

Thanks for reporting this 👍

PR #83, opened 2021-11-19
Forbid RBI outside of configurable AllowedPaths

RyanBrushett:
https://github.com/Shopify/rubocop-sorbet/issues/79

`ForbidRBIOutsideOfSorbetDir` makes the assumption that RBI files will always be kept under `sorbet/rbi` which is not necessarily the case. [See the sorbet docs on RBI files within gems](https://sorbet.org/docs/rbi#rbis-within-gems) for one example of this. Refactoring the cop a bit allows us to be flexible for these sorts of setups.

AllowedPaths will default to `["sorbet/rbi/**"]` to be consistent with `ForbidRBIOutsideOfSorbetDir`
This cop makes no attempts to tell a user where to put things, but allows them to tailor the cop to their needs.
I decided that having AllowedPaths set explicitly to an empty list, nil value, or list containing nil meant it was broken.

See tests for a rundown.


lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
```diff
@@ -0,0 +1,47 @@
...
+      #   # sorbet/rbi/any/path/for/file.rbi
+      class ForbidRBIOutsideOfAllowedPaths < RuboCop::Cop::Cop
+        include RangeHelp
+
+        def investigate(processed_source)
+          add_offense(
+            nil,
+            location: source_range(processed_source.buffer, 1, 0),
+            message: message
+          ) if allowed_paths.none? { |pattern| File.fnmatch(pattern, processed_source.file_path) }
```
vinistock, line 28: TIL about `fnmatch`.

vinistock (approved):

PR #89, opened 2021-12-07
Fix ForbidRBIOutsideOfAllowedPaths when paths are absolute

Morriar:
Three fixes in this PR:

1. Better handling of bad values passed to `AllowedPaths`: the cop was crashing when the configuration contained: `AllowedPaths: sorbet/rbi/**` because we were not expecting a string here
2. Fix a type in a test path
3. Handle file with absolute paths: the cop was creating an offence for `/home/my/project/sorbet/rbi/shims/foo.rbi` when the config was `sorbet/rbi/**` while it shouldn't

cc. @RyanBrushett


lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
```diff
@@ -21,25 +23,41 @@ class ForbidRBIOutsideOfAllowedPaths < RuboCop::Cop::Cop
...
+              nil,
+              location: source_range(processed_source.buffer, 1, 0),
+              message: "AllowedPaths cannot be empty"
+            )
+            return
+          end
+
+          # When ran with `bundle exec` the path to the source file is absolute.
+          # We need to remove the exec path directory prefix before matching with the filename regular expressions.
+          rel_path = processed_source.file_path.sub("#{Dir.pwd}/", "")
```
Morriar, line 46: I'm not super happy about this fix but I don't have a better idea on how to find the relative path to the file from the project root.

My bet is that the paths in the config should match the directory from where you expect to run your rubocop command.

Morriar: > Is the path to the file only absolute when running with `bundle exec`?

I double checked and no, it's actually the same with or without `bundle exec`.

> I'm wondering if it would simplify things to take the `allow_paths` and manipulate them instead.

How would you go about it? Using `"#{Dir.pwd}/#{pattern}"` instead of a bare `pattern`? It seems equivalent.

vinistock: Yeah, I guess so. Another idea is using `Dir` and `Pathname` to get all files that match with the complete path and check if it's included in the list. Something like

```ruby
all_matches = Dir[pattern]
full_paths = all_matches.map { |path| Pathname.new(path).realpath } # full_paths is now a list of all files that match the pattern with the full path
```

But I guess it might be slower and not as elegant as our previous approach.

RyanBrushett: > I double checked and no, it's actually the same with or without bundle exec.

Sad I missed this tbh.
Can you update the comment?

RyanBrushett (approved):
Good eye 🤦 Should have tested on a real-life system. Duly noted.
Thank you!

vinistock (approved):

PR #92, opened 2022-02-08
Add TypeAliasName to ensure all constants used as T.type_alias are in CamelCase

Morriar:
This cop ensures all constants used as `T.type_alias` are using CamelCase:
     
~~~rb
# bad
FOO_OR_BAR = T.type_alias { T.any(Foo, Bar) }

# good
FooOrBar = T.type_alias { T.any(Foo, Bar) }
~~~

### DIFF
--- a/lib/rubocop/cop/sorbet/type_alias_name.rb
+++ b/lib/rubocop/cop/sorbet/type_alias_name.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop ensures all constants used as `T.type_alias` are using CamelCase.
+      #
+      # @example
+      #
+      #   # bad
+      #   FOO_OR_BAR = T.type_alias { T.any(Foo, Bar) }
+      #
+      #   # good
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/sorbet/type_alias_name.rb
```diff
@@ -0,0 +1,45 @@
...
+            ))
+        PATTERN
+
+        def on_casgn(node)
+          return unless casgn_type_alias?(node)
+
+          name = node.children[1]
+
+          # From https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/naming/class_and_module_camel_case.rb
+          return unless /_/.match?(name)
```
RyanBrushett, line 38: I think this will pass happily if `name` is something like `"HELLO"`

Should we also check for `name == name.capitalize`?

vinistock: That's fair, I agree with only marking names with underscores as offenses.

vinistock (approved):

RyanBrushett (approved):

KaanOzkan (approved):
lgtm after typo

PR #101, opened 2022-03-08
Refine `Style/MutableConstant` to process `T.let` expressions.

paracycle:
The built-in `Style/MutableConstant` cop only catches naked assignments to constants and missed the assignments that are wrapped in `T.let` expressions. By simply refining `Style/MutableConstant` to process any `T.let` expression, we can catch all assignments and do the correct processing even in the presence of Sorbet.

There was a previous attempt in #36 to reimplement `Style/MutableConstant` which was later rejected since we didn't want to keep maintaining the whole cop inside this repo. To wit, the cop changed drastically upstream.

However, we can still decorate the upstream cop with a prepended module that further processes the matched node through a `T.let` expression matcher before doing its magic.

I've mostly copy/pasted the upstream tests and modified them to use `T.let` in test cases. However, in shared contexts, I've also included non-`T.let` usages too, to make sure we were not breaking upstream behaviour.

### DIFF
--- a/manual/cops.md
+++ b/manual/cops.md
@@ -34,3 +34,8 @@ In the following section you find all available cops:
 * [Sorbet/ValidSigil](cops_sorbet.md#sorbetvalidsigil)
 
 <!-- END_COP_LIST -->
+
+In addition to the cops defined in this gem, it also modifies the behaviour of some other cops
+defined in other RuboCop gems:
+
+* [Style/MutableConstant](https://docs.rubocop.org/rubocop/cops_style.html#stylemutableconstant): In addition to the default behaviour, RuboCop Sorbet makes this cop `T.let` aware, so that `CONST = T.let([1, 2, 3], T::Array[Integer])` is also treated as a mutable literal constant value.


lib/rubocop/cop/sorbet/mutable_constant_sorbet_aware_behaviour.rb
```diff
@@ -0,0 +1,29 @@
...
+          super(value)
+        end
+      end
+    end
+  end
+end
+
+RuboCop::Cop::Style::MutableConstant.prepend(
+  RuboCop::Cop::Sorbet::MutableConstantSorbetAwareBehaviour
+)
```
Morriar, line 29: Can we move this inside `RuboCop::Cop::Sorbet` so it fits on one line?

Morriar (approved):

lib/rubocop/cop/sorbet/mutable_constant_sorbet_aware_behaviour.rb
paracycle: That's how I had it first, but it felt like the fact that we are modifying the behaviour of a class that we are NOT defining was being lost inside that context, since there are many other things going on in there.

The explicit prepend on its own felt way better to me.

PR #107, opened 2022-05-05
Fix BindingConstantWithoutTypeAlias for type members with arguments

Morriar:
This shouldn't be an offense:

```
Sorbet/BindingConstantWithoutTypeAlias: It looks like you're trying to bind a type to a constant.
To do this, you must alias the type using T.type_alias.
--
  | A = type_member(:in) { { fixed: T.untyped } }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```


paracycle (approved):
Thank you!

PR #109, opened 2022-06-16
Do not search for sigils in the middle of comments

rafaelfranca:
If you have a comment mentioning a sigil, the EnforceSingleSigil cop was failing. This should not happen.


st0012 (approved):

Morriar (approved):
Amazing! 😂

RyanBrushett (approved):
Thanks for this!

PR #111, opened 2022-06-16
Disable Sorbet/ForbidTUntyped by default

rafaelfranca:
It was enabled by default by mistake. Disabling now and updating which version added it.


RyanBrushett:
Duplicate of #110

Morriar:
I think it's a duplicate of https://github.com/Shopify/rubocop-sorbet/pull/110.

Issue #115, opened 2022-07-21
`ForbidUntypedStructProps` does not work with `T::ImmutableStruct`


ghiculescu:
https://github.com/Shopify/rubocop-sorbet/blob/main/lib/rubocop/cop/sorbet/forbid_untyped_struct_props.rb needs support for https://github.com/sorbet/sorbet/pull/6005

PR #116, opened 2022-07-21
Support `T::ImmutableStruct` in `ForbidUntypedStructProps`

ghiculescu:
Fixes https://github.com/Shopify/rubocop-sorbet/issues/115


KaanOzkan (approved):
Thanks. It may be worth to add `< T::Struct` to the cop documentation [comment](https://github.com/Shopify/rubocop-sorbet/blob/8e094ecc0ac8f487b745378c9f4133d3170500b8/lib/rubocop/cop/sorbet/forbid_untyped_struct_props.rb#L15) as well now that you're editing the file.

KaanOzkan (approved):

PR #120, opened 2022-08-12
Update documentation for sigil order cop

ameketa:
The EnforceSigilOrder cop originally preferred to put the `typed` sigil
first, above the `encoding` sigil, which caused Ruby to ignore the
`encoding` sigil if present. This was corrected in:

66b2d6f48b88d7f290727c7745584ca830ab3316

but the documentation was not updated in that same commit.

This change updates the documentation in the cop to reflect the correct
ordering of the sigils enforced by the cop.


Morriar (approved):
Good catch, thanks!

ameketa:
I just signed the CLA as well!

PR #121, opened 2022-08-12
Update documentation

ameketa:
The documentation for the cop that forbids untyped struct props was updated in:

3456b3e72adc431373cd1a0fec684b39711780e5

but the associated generated documentation wasn't updated at the same time.

This change includes the documentation changes in the documents that get generated via `bundle exec rake generate_cops_documentation`.


ameketa:
(noticed in my previous PR that updating documentation pulled in these changes, too, so wanted to get these in separately!)

paracycle:
Ugh, we should really check that the documentation is up to date on CI. Thanks for your contribution.

PR #127, opened 2023-01-17
Add rbi/** to default AllowedPaths for ForbidRBIOutsideOfAllowedPaths

dduugg:
The `ForbidRBIOutsideOfAllowedPaths` creates an offense when following the [official guidance](https://sorbet.org/docs/rbi#rbis-within-gems) on where to place `rbi` files within gem code:

> We expect that as adoption grows, gems will include their external interfaces RBI files in an `rbi/` directory. When they do this, Tapioca will automatically merge those definitions in the autogenerated RBIs. In the future, we anticipate this to be the preferred way to include RBI files into a project.

This was [confirmed](https://sorbet-ruby.slack.com/archives/C02VD06EQ3U/p1673923210096859?thread_ts=1673917916.215189&cid=C02VD06EQ3U) via the Sorbet Slack with a pointer to the tapioca [code](https://github.com/Shopify/tapioca/blob/1b485c6ce47fb840554fbb864d58ade1e2be20df/lib/tapioca/gemfile.rb#L217-L220).

Since it's confusing for end-users following the docs to trigger violations, I propose including this path in the default config. (Also, the cop message doesn't mention `AllowedPaths`, adding to the confusion.)

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -78,6 +78,7 @@ Sorbet/ForbidRBIOutsideOfAllowedPaths:
   Enabled: true
   VersionAdded: 0.6.1
   AllowedPaths:
+  - "rbi/**"
   - "sorbet/rbi/**"
   Include:
   - "**/*.rbi"
--- a/lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
+++ b/lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
@@ -9,7 +9,7 @@ module RuboCop
       #
       # Options:
       #
-      # * `AllowedPaths`: A list of the paths where RBI files are allowed (default: ["sorbet/rbi/**"])
+      # * `AllowedPaths`: A list of the paths where RBI files are allowed (default: ["rbi/**", "sorbet/rbi/**"])
       #
       # @example
       #   # bad
@@ -17,6 +17,7 @@ module RuboCop
       #   # other_file.rbi
       #
       #   # good
+      #   # rbi/external_interface.rbi
       #   # sorbet/rbi/some_file.rbi
       #   # sorbet/rbi/any/path/for/file.rbi
       class ForbidRBIOutsideOfAllowedPaths < RuboCop::Cop::Cop
--- a/manual/cops_sorbet.md
+++ b/manual/cops_sorbet.md
@@ -311,7 +311,7 @@ This cop makes sure that RBI files are always located under the defined allowed 
 
 Options:
 
-* `AllowedPaths`: A list of the paths where RBI files are allowed (default: ["sorbet/rbi/**"])
+* `AllowedPaths`: A list of the paths where RBI files are allowed (default: ["rbi/**", "sorbet/rbi/**"])
 
 ### Examples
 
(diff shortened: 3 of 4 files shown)


Morriar:
:wave: Hey @dduugg,

I'm not sure this is the right approach and I would like to get some opinions before merging this.

The `rbi/` directory for gems pre-dates [Tapioca becoming the default RBI generator for Sorbet](https://sorbet.org/blog/2022/07/27/srb-tapioca).

Before, the recommended way to share RBIs from a gem was indeed to put them in the `rbi/` directory. However, maintaining the `rbi/` directory up to date with the actual code was painful and most of the time the RBIs shared that way were actually wrong. The `rbi/` directory also has the limitation of [not working with LSP mode](https://github.com/sorbet/sorbet/issues/2496).

Thus, we decided on another approach for Tapioca: [it imports the type annotations it finds in the gem's code](https://github.com/Shopify/tapioca/pull/61) directly and merge them into the generated RBIs for the gem. This way you can type your gem with Sorbet and validate the types used with Sorbet. No more outdated signatures and a better experience for the end user.

We also made a change in Sorbet to ignore the `rbi/` directory if Tapioca is present in the Gemfile. Some context here: https://github.com/sorbet/sorbet/pull/3332.

Now, for retro-compatibility, Tapioca still look into the `rbi/` directory and [tries to merge its content into the generated RBIs](https://github.com/Shopify/tapioca/pull/453). So it _could_ make sense to allow them with this cop.

Since Tapioca is now the official way of generating RBIs for Sorbet, I think we should encourage people to rely on signatures inside their gem's code rather than in the `rbi/` directory. So maybe the right fix would be to stop recommending using `rbi/` or at least make it less encouraged in the Sorbet documentation. Maybe we could also document how to allow it in our rubocop configuration better if you still want to use the `rbi/` directory.

What do you think?

dduugg:
Hi @Morriar, thanks for your thoughtful reply. I'll expand on my use case, in case it's helpful:

I have a gem that ships a tiny `.rbi` file: https://github.com/dduugg/yard-sorbet/blob/main/lib/yard-sorbet/handlers/struct_class_handler.rbi

The reason for this is that I don't want the runtime to depend on an [experimental](https://sorbet.org/docs/requires-ancestor) sorbet feature. The file currently lives in `lib`, alongside the class it decorates, but `rubocop-sorbet` suggests moving it to `sorbet/rbi`. I then noticed the contradiction with sorbet docs and filed this ticket.

One additional consideration is that there are some utilities that wish to include type signatures but not a `sorbet-runtime` dependency. One example is [pdf-reader](https://github.com/yob/pdf-reader), which [ships](https://github.com/yob/pdf-reader/blob/main/pdf-reader.gemspec#L9) with an `rbi/` dir as the sorbet docs recommends, but does not use sorbet-runtime. (The maintainer also has a [blog post](https://yob.id.au/2021/12/31/sorbet-with-gems.html) that expands on this approach.)

Your point about keeping `rbi/` in sync is something that I've [seen](https://github.com/Homebrew/brew/pull/14364) as well, but I think based on the above, there may be some corner cases where type annotations outside of the runtime code is necessary/desirable, and merit continued support. Whether that's alongside the code (`sorbet-runtime`, `brew`), in `rbi/` (`pdf-reader`), or in `sorbet/rbi` (`tapioca`), isn't especially important IMO (though the first option slightly reduces the chance of drift), just something that should have clear guidance from the sorbet docs and tooling.

Hope this context helps, and thank you for your work on `rubocop-sorbet`.

Morriar:
> One additional consideration is that there are some utilities that wish to include type signatures but not a sorbet-runtime dependency. 

Yeah this is a very good point!

PR #131, opened 2023-02-02
Add new `Sorbet/RedundantExtendTSig` cop

sambostock:
This cop is intended to be used in applications that have monkey patched `T::Sig` into all modules.

```ruby
Module.include(T::Sig)

# later
module MyModule
  extend T::Sig
# ^^^^^^^^^^^^^ Do not redundantly `extend T::Sig` when it is already included in all modules.
  sig { void }
  def whatever = nil
end
```

It is disabled by default, as only application developers themselves can know if it applies to their app.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -156,6 +156,16 @@ Sorbet/OneAncestorPerLine:
   Enabled: false
   VersionAdded: '0.6.0'
 
+Sorbet/ForbidExtendTSig:
+  Description: >-
+    Forbid the usage of redundant `extend T::Sig`.
+
+    Only for use in applications that monkey patch `Module.include(T::Sig)` directly,
+    where it is useful to reduce noise.
+  Enabled: false
+  Safe: false
+  VersionAdded: '<<next>>'
+
 Sorbet/SignatureBuildOrder:
   Description: >-
                   Enforces the order of parts in a signature.
--- a/lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
+++ b/lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Forbids the use of redundant `extend T::Sig`. Only for use in
+      # applications that monkey patch `Module.include(T::Sig)` globally,
+      # which would make it redundant.
+      #
+      # @safety
+      #   This cop should not be enabled in applications that have not monkey
+      #   patched `Module`.
+      #
+      # @example
+      #   # bad
+      #   class Greeter
+      #     extend T::Sig
+      #     sig { void }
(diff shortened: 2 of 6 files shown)


lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
```diff
@@ -0,0 +1,48 @@
...
+      #   end
+      #
+      #   # good
+      #   class Greeter
+      #     sig { void }
+      #     def say_hello = puts "hello"
+      #   end
+      #
+      class RedundantExtendTSig < Base
+        extend AutoCorrector
```
sambostock, line 29: Looks like the way we generate docs doesn't support the `Base` API, and requires the old `Cop` API.  I've updated it to the legacy approach to correction, and will open an issue to migrate.

KaanOzkan (commented):
I like the cop 👍 . You have a CI failure on Ruby 3+

sambostock:
Switched to positional arguments. I'll squash commits when I'm back at my computer.

What do you think of the offense message/docs?

lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Forbids the use of redundant `extend T::Sig`. Meant for use in
```
paracycle, line 6: Can we instead say:

```suggestion
      # Forbids the use of redundant `extend T::Sig`. Only for use in
```

config/default.yml
```diff
@@ -156,6 +156,16 @@ Sorbet/OneAncestorPerLine:
   Enabled: false
   VersionAdded: '0.6.0'
 
+Sorbet/ForbidExtendTSig:
+  Description: >-
+    Forbid the usage of redundant `extend T::Sig`.
+
+    For use in applications that monkey patch `Module.include(T::Sig)` directly,
```
paracycle, line 163: Similarly:

```suggestion
    Only for use in applications that monkey patch `Module.include(T::Sig)` directly,
```

paracycle (approved):
Small wording change to better emphasize which apps should use the cop.

vinistock (approved):

config/default.yml
```diff
@@ -156,6 +156,16 @@ Sorbet/OneAncestorPerLine:
...
+    where it is useful to reduce noise.
+  Enabled: false
+  Safe: false
+  VersionAdded: '<<next>>'
```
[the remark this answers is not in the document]
sambostock: I assume Rubocop proper has tooling that does this automatically during the release process, but we haven't copied their tooling exactly (see #132). I think the expected workflow would be for the PR that bumps the version to do a find and replace of `<<next>>` with whatever version it's bumping the gem to.

PR #136, opened 2023-02-03
Add Ruby 3.2 to test matrix & use it in development

sambostock:
This adds Ruby 3.2 to the test matrix & uses it in development. This doesn't change anything else. RuboCop remains configured to enforce 2.5 compatible syntax.

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -3,7 +3,7 @@ name: "rubocop-sorbet"
 type: "ruby"
 
 up:
-  - ruby: "2.5.5"
+  - ruby: "3.2.0"
   - "bundler"
 
 commands:


andyw8 (approved):

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1' ]
+        ruby: [ '2.5', '2.6', '2.7', '3.0', '3.1', '3.2' ]
```
andyw8, line 14: I wonder if we should drop 2.5 and 2.6.

andyw8: RuboCop itself tests against 2.6 upwards:

https://github.com/rubocop/rubocop/blob/master/.circleci/config.yml

vinistock (approved):

.github/workflows/ci.yml
sambostock: The upcoming end-of-life of Ruby 2.7 might be a good time to drop everything older than Ruby 3, although that might be a little aggressive.

PR #138, opened 2023-02-03
Fix default config for `RedundantExtendTSig`

sambostock:
In the original draft, the cop was called `ForbidExtendTSig`. It was later renamed, and the config was moved to what would be the correct location alphabetically, but the actual config key was not updated.

This cause the cop to have no default config, so it was enabled by default, which would have been problematic for consumers.

### DIFF
--- a/manual/cops_sorbet.md
+++ b/manual/cops_sorbet.md
@@ -501,7 +501,7 @@ end
 
 Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 --- | --- | --- | --- | ---
-Enabled | Yes | Yes  | - | -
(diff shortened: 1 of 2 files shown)


manual/cops_sorbet.md
```diff
@@ -501,7 +501,7 @@ end
 
 Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
 --- | --- | --- | --- | ---
-Enabled | Yes | Yes  | - | -
```
sambostock, line 504: This was missed in #131, but should have been a giveaway that something was wrong.

vinistock (approved):

PR #139, opened 2023-02-03
Fix missing `sig` false positive on `sig(:final)`

### LINKED ISSUE #134: `EnforceSignatures` cop doesn't detect `sig(:final)`
It appears that the `EnforceSignatures` cop has trouble accepting `sig(:final)` as a valid signature. This might also affect other signature-based cops.

Here's a failing code example.
```ruby
# typed: true

class Boundary
  sig(:final) { returns(Elem) }
  def call
    trace { filter }
  end
end
```

Here's the cop configuration:
```yml
Sorbet/EnforceSignatures:
  Enabled: true
  AutoCorrect: false
  Severity: warning
  Exclude:
    - 'config/**/*'
   ...
```

imagen

#### Expected behaviour:

The method doesn't trigger the cop's warning, since it has a valid signature.

sambostock:
Closes #134, and makes a slight refactor along the way.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/signature_cop.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_cop.rb
@@ -15,6 +15,7 @@ module RuboCop
           (block (send
             {nil? #with_runtime? #without_runtime?}
             :sig
+            (sym :final)?
           ) (args) ...)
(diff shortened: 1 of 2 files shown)


vinistock (approved):
Nice 🙏

PR #141, opened 2023-02-18
Add `Sorbet/FetchWhenMust` cop

sambostock:
Checks for `T.must(object[key])` and recommends `object.fetch(key)` instead.

While usually safe, false positives are possible if `object` does not respond to `fetch`, or if `Hash#default_proc` (or similar) is being used.

### DIFF
--- a/lib/rubocop/cop/sorbet/fetch_when_must.rb
+++ b/lib/rubocop/cop/sorbet/fetch_when_must.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # rubocop:disable Lint/RedundantCopDisableDirective
+
+      # Checks for `T.must(object[key])` and recommends `object.fetch(key)` instead.
+      #
+      # @safety
+      #   False positives are possible if `object` does not respond to `fetch`,
+      #   or if `Hash#default_proc` (or similar) is being used.
+      #
+      # @example
+      #   # bad
+      #   T.must(object[key])
+      #
+      #   # good
+      #   object.fetch(key)
+      #
+      #   # good
+      #   object[key]
+      #
+      #   # good
+      #   # If `object` does not `respond_to? :fetch`, or if using `Hash` `default_proc`
+      #   T.must(object[key]) # rubocop:disable Sorbet/FetchWhenMust
+      #
+      class FetchWhenMust < RuboCop::Cop::Base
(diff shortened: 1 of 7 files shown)


andyw8:
I'd love to have this, but I worry there be some pushback since `fetch` is not commonly used on arrays.

sambostock:
I discovered an issue with nested offense correction where correcting the inner node would clobber the outer node correction.

After investigating, I found rubocop/rubocop#10511 & rubocop/rubocop#10461 which demonstrate how to address this skipping correction for nodes within an already corrected node. Because autocorrection repeats investigation until the file no longer requires correction, all nodes are eventually corrected. I have taken this approach here.

amomchilov (approved):
This is a common thing I suggest on PRs. Love the idea!

lib/rubocop/cop/sorbet/fetch_when_must.rb
```diff
@@ -0,0 +1,69 @@
...
+      #
+      #   # good
+      #   object.fetch(key)
+      #
+      #   # good
+      #   object[key]
+      #
+      #   # good
+      #   # If `object` does not `respond_to? :fetch`, or if using `Hash` `default_proc`
+      #   T.must(object[key]) # rubocop:disable Sorbet/FetchWhenMust
```
Morriar, line 26: I'm not completely convinced by this cop exactly because of this case.

From a user experience point of view:

1. I write some code that is seemingly correct
2. The cop complains that I should use `fetch`
3. I try to use fetch but now the type checker complains
4. What should I do? Adding a `rubocop:disable` always feels wrong...
5. Confusion state

I agree that is this something we generally recommend in code reviews but the inability to only target hashes / arrays from the cop is a huge limitation in this case.

Do we have an idea of how many `rubocop:disable` we would need to add in core for example?

I'm also worried about the massive change needed to enable this cop. What if the Sorbet nor tests do not catch that one of the `fetch` call is actually invalid?

amomchilov: Is there a way to leverage Sorbet's type information to narrow this cop to only Arrays/Hashes and other similar container types?

andyw8:
Minor data point – a Stripe engineer suggests `Array#fetch` here:

https://sorbet-ruby.slack.com/archives/CHN2L03NH/p1686444818750959?thread_ts=1686410073.163849&cid=CHN2L03NH

> Sorbet does not track the information you want here. if you want a non-nilable type, you should use `Array#fetch`.

PR #148, opened 2023-02-21
Refactor `ForbidUntypedStructProps`

sambostock:

- Consolidates `T::Struct` and `T::ImmutableStruct` node matchers.
- Use a capture in node search instead of indexing into child nodes.


lib/rubocop/cop/sorbet/forbid_untyped_struct_props.rb
```diff
@@ -47,19 +42,19 @@ class ForbidUntypedStructProps < RuboCop::Cop::Base
...
-          (send nil? {:prop :const} _ {#t_untyped #t_nilable_untyped} ...)
+          (send nil? {:prop :const} _ ${#t_untyped #t_nilable_untyped} ...)
         PATTERN
 
         def on_class(node)
           return unless subclass_of_t_struct?(node)
 
-          untyped_props(node).each do |untyped_prop|
-            add_offense(untyped_prop.child_nodes[1])
+          untyped_props(node).each do |prop_type|
```
paracycle, line 56: Do you mind adding a comment here or on top of `def_node_search :untyped_props` that says that the `${ ... }` part in the node matcher is what captures the `prop_type` here.

I like the terseness of the usage, but it might be hard to understand what's going on for someone new to the project.

lib/rubocop/cop/sorbet/forbid_untyped_struct_props.rb
```diff
@@ -47,19 +42,19 @@ class ForbidUntypedStructProps < RuboCop::Cop::Base
 
         # @!method subclass_of_t_struct?(node)
         def_node_matcher :subclass_of_t_struct?, <<~PATTERN
-          (class (const ...) {#t_struct #t_immutable_struct} ...)
+          (class (const ...) #t_struct ...)
         PATTERN
 
         # @!method untyped_props(node)
```
sambostock, line 48: @paracycle what do you think of this (RE: https://github.com/Shopify/rubocop-sorbet/pull/148/files#r1113271304)

```suggestion
        # @!method untyped_props(node)
        # Search for untyped prop/const declarations and capture their types
```

paracycle: Ok, that's fair. Let's go with your suggested change.

sambostock: Okay, sounds good. I've opened https://github.com/Shopify/rubocop-sorbet/pull/155 (although its currently empty) for us to fill in contributing documentation, where we can link to the node pattern docs as something good to know before getting started.

paracycle (approved):

PR #153, opened 2023-02-21
Refactor `ForbidIncludeConstLiteral`

sambostock:

- Migrates from deprecated `Cop` parent class to `Base`
- Improves the offense message
- Removes the unused `used_names` accessor and initialization
- Renames and alters the node matcher to describe what it matches, not what it doesn't
- Extracts named helpers to clarify the code's logic
- Leverages the `expect_correction` test helpers and replaces the autocorrection specific tests.


lib/rubocop/cop/sorbet/forbid_include_const_literal.rb
```diff
@@ -26,41 +26,44 @@ module Sorbet
...
-            )
-          end
+        private
+
+        def neither_const_nor_self?(node)
+          !node.const_type? && !node.self_type?
+        end
+
+        # Returns true if the node is within a module declaration that is not anonymous.
+        def within_onymous_module?(node)
```
paracycle, line 63: Do you mean:
```suggestion
        def within_anonymous_module?(node)
```

lib/rubocop/cop/sorbet/forbid_include_const_literal.rb
```diff
@@ -26,41 +26,44 @@ module Sorbet
...
 
         def on_send(node)
-          return unless not_lit_const_include?(node) do |send_argument|
-            ![:const, :self].include?(send_argument.type)
+          dynamic_inclusion?(node) do |receiver, inclusion_method, included|
+            return if t_unsafe?(receiver)
+            return unless within_onymous_module?(node)
+
+            add_offense(node, message: format(MSG, inclusion_method: inclusion_method)) do |corrector|
+              explicit_receiver = receiver&.source || 'self'
```
paracycle, line 50: This seems to be a bug-fix so I think we have no test cases for the explicit receiver not being `self`. Can we add one or two cases for that as well?

spec/rubocop/cop/sorbet/forbid_include_const_literal_spec.rb
```diff
@@ -64,12 +101,54 @@ module FilterHelper
...
+  it "adds offense when an explicit constant receiver prepends a send node" do
+    expect_offense(<<~RUBY)
+      MyClass.prepend Rails.application.routes.url_helpers
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `prepend` must only be used with constant literals as arguments
+    RUBY
+
+    expect_correction(<<~RUBY)
+      T.unsafe(MyClass).prepend Rails.application.routes.url_helpers
+    RUBY
+  end
```
sambostock, line 150: @paracycle, I've added these tests as per https://github.com/Shopify/rubocop-sorbet/pull/153#discussion_r1113640476.

lib/rubocop/cop/sorbet/forbid_include_const_literal.rb
sambostock: I've added tests, although I should note that they look like

```ruby
module M
  M.include(dynamic)
end
```

which this implementation detects, not

```ruby
M.include(dynamic)
```

which this implementation does _not_ detect. Do we want to do that? I'm unsure about false positives.

paracycle (approved):

lib/rubocop/cop/sorbet/forbid_include_const_literal.rb
Morriar: I think we can tune down on the sesquipedalianism and keep the codebase easy to understand without a dictionary. 

`within_named_module?` would have worked.

PR #154, opened 2023-02-21
Refactor `BindingConstantWithoutTypeAlias`

sambostock:

- Renames the file to match the cop name
  - Because of this, and the number of changes made, Git considers this an entirely new file. 🤷
- Migrates from the deprecated `Cop` parent class to `Base`
- Uses idiomatic `MSG` constant
- Reduces the number of matchers and negations
- Uses positive naming, rather then negative naming
- Uses captures instead of indexing children, when possible
- Uses named helpers when navigating AST
- Uses `expect_corrections` to improve tests


lib/rubocop/cop/sorbet/binding_constant_without_type_alias.rb
```diff
@@ -0,0 +1,102 @@
...
+              :self_type
+              :untyped
+            }
+            ...
+          )
+        PATTERN
+
+        def on_casgn(node)
+          expression = node.expression
+          return if expression.nil? # multiple assignment
```
sambostock, line 64: Technically someone could do

```ruby
A, B = T.any(Some, Thing), T.all(Something, Else)
```

which we'd want to flag, and autocorrect to

```ruby
A, B = T.type_alias { T.any(Some, Thing) }, T.type_alias { T.all(Something, Else) }
```

or more likely

```ruby
A = T.type_alias { T.any(Some, Thing) }
B = T.type_alias { T.all(Something, Else) }
```

However, it seems very unlikely anyone would do actually do this, so I'm not sure it's worth the complexity to try to detect it.

lib/rubocop/cop/sorbet/binding_constant_without_type_alias.rb
```diff
@@ -0,0 +1,102 @@
...
+        #
+        # i.e. in AST
+        #
+        #     (send (send (send (send (send (send nil :a) :b) :c) :d) :e) :f)
+        #                             ^^^^^^^^^^^^^^^^^^^^^^^
+        #
+        def send_leaf(node)
+          node = node.receiver while node&.receiver&.send_type?
+          node
+        end
```
sambostock, line 98: This was hard to name. I'm still not satisfied with it, but it's the best I've got right now.

I feel like the concept is simple, but the fact that the Ruby code and AST are "inverted" makes the root/leaf/parent terminology awkward.

Long story short

```ruby
T.proc.void
^^^^^^ extracts this
```

I made it a loop, so that it will continue to work even if there's some weird usage like

```ruby
T.proc.foo.bar.baz
```

paracycle: How come we didn't need this in the old implementation?

paracycle (approved):
This looks good to me.

lib/rubocop/cop/sorbet/binding_constant_without_type_alias.rb
sambostock: The old implementation was using [this node search](https://github.com/Shopify/rubocop-sorbet/blob/0333a8e25e06890dd100d8313f06ad4995a57cac/lib/rubocop/cop/sorbet/binding_constants_without_type_alias.rb#L79-L87), which meant it would traverse all descendant looking for a match anywhere.

This means it would get confused by the following

```ruby
A = lambda do
  # ^^^^^^^^^ It looks like you're trying to bind a type to a constant. To do this, you must alias the type using `T.type_alias`.
  x = T.let(1, T.untyped) # Because it erroneously detects this `T.untyped`
  x
end
```

while that's unlikely, the new implementation tries to be more explicit about what it's looking for: we only care about the starting `T.something` method send.

vinistock (approved):

PR #162, opened 2023-06-01
Add cop to rewrite obsolete memoization pattern

amomchilov:
Until recently, memoizing with instance variables in `#typed: strict` files required this cumbersome pattern:

```ruby
sig { returns(Foo) }
def foo
  @foo = T.let(@foo, T.nilable(Foo))
  @foo ||= Foo.new
end
```

This was fixed in Sorbet `0.5.10210.20220725105033-6e80c502e` (see https://github.com/sorbet/sorbet/issues/995)! 🥳 

This new `Sorbet/ObsoleteStrictMemoization` cop catches that obsolete memoization pattern, and rewrites it into:

```ruby
sig { returns(Foo) }
def foo
  @foo ||= T.let(Foo.new, T.nilable(Foo))
end
```

The rewriter also has some special logic to make sure that the resulting line isn't too long. If it is, it'll automatically split it into multiple lines. Look at the test cases for more details.

#### TODO

1. [x] Detect offences even if they're not in the first 2 lines of a method

    There's a limitation in this cop where it can't detect this pattern unless it's the first thing in a method body. Fixing this seemed pretty tricky, so I left it out for now.

    Some of these missed offences of this pattern can be found with this regex:

    ```regex
    \@([a-zAZ_]+) = T\.let\(@\1
    ```

2. [x] Improved the way indentation rules are detected and applied
3. [x] Only apply when the user's Sorbet version is greater than `0.5.10210.20220725105033-6e80c502e`

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -159,6 +159,18 @@ Sorbet/KeywordArgumentOrdering:
   Enabled: true
   VersionAdded: 0.2.0
 
+Sorbet/ObsoleteStrictMemoization:
+  Description: >-
+    This cop checks for the obsolete pattern for initializing instance variables that was required for older Sorbet
+    versions in `#typed: strict` files.
+
+    It's no longer required, as of Sorbet 0.5.10210
+    See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization
+  Enabled: true
+  VersionAdded: '0.7.1'
+  Safe: true
+  SafeAutoCorrect: true
+
 Sorbet/OneAncestorPerLine:
   Description: 'Enforces one ancestor per call to requires_ancestor'
   Enabled: false
--- a/lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
+++ b/lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
@@ -29,22 +29,18 @@ module RuboCop
         end
 
         def target_sorbet_static_version_from_bundler_lock_file
-          @target_sorbet_static_version_from_bundler_lock_file ||= read_sorbet_static_version_from_bundler_lock_file
+          if defined?(@target_sorbet_static_version_from_bundler_lock_file)
+            @target_sorbet_static_version_from_bundler_lock_file
+          else
+            @target_sorbet_static_version_from_bundler_lock_file = read_sorbet_static_version_from_bundler_lock_file
+          end
         end
 
-        # Adapted from https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/config.rb#L293-L308
         def read_sorbet_static_version_from_bundler_lock_file
-          lock_file_path = config.bundler_lock_file_path
-
-          return nil unless lock_file_path
-
-          File.foreach(lock_file_path) do |line|
-            # If Sorbet (or one of its frameworks) is in Gemfile.lock or gems.lock, there should be
-            # a line like:
-            #         sorbet-static (X.X.X-some_arch)
-            result = line.match(/^\s+sorbet-static\s+\((\d+\.\d+\.\d+)/)
-            return Gem::Version.new(result.captures.first) if result
-          end
+          require "bundler"
+          ::Bundler.locked_gems.specs.find { |spec| spec.name == "sorbet-static" }&.version
+        rescue LoadError, Bundler::GemfileNotFound
+          nil
         end
       end
     end
--- a/lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
+++ b/lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      module TargetSorbetVersion
+        class << self
+          def included(target)
+            target.extend(ClassMethods)
+          end
+        end
+
+        module ClassMethods
+          # The version of the Sorbet static type checker required by this cop
+          def minimum_target_sorbet_static_version(version)
+            @minimum_target_sorbet_static_version = Gem::Version.new(version)
+          end
+
+          def support_target_sorbet_static_version?(version)
+            @minimum_target_sorbet_static_version <= Gem::Version.new(version)
+          end
+        end
+
+        def enabled_for_sorbet_static_version?
+          sorbet_static_version = target_sorbet_static_version_from_bundler_lock_file
+          return false unless sorbet_static_version
+
+          self.class.support_target_sorbet_static_version?(sorbet_static_version)
+        end
+
+        def target_sorbet_static_version_from_bundler_lock_file
+          @target_sorbet_static_version_from_bundler_lock_file ||= read_sorbet_static_version_from_bundler_lock_file
+        end
+
+        # Adapted from https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/config.rb#L293-L308
+        def read_sorbet_static_version_from_bundler_lock_file
+          lock_file_path = config.bundler_lock_file_path
+
+          return nil unless lock_file_path
+
+          File.foreach(lock_file_path) do |line|
+            # If Sorbet (or one of its frameworks) is in Gemfile.lock or gems.lock, there should be
+            # a line like:
+            #         sorbet-static (X.X.X-some_arch)
+            result = line.match(/^\s+sorbet-static\s+\((\d+\.\d+\.\d+)/)
+            return Gem::Version.new(result.captures.first) if result
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
+++ b/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
@@ -1,0 +1,106 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Checks for the obsolete pattern for initializing instance variables that was required for older Sorbet
+      # versions in `#typed: strict` files.
+      #
+      # It's no longer required, as of Sorbet 0.5.10210
+      # See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization
+      #
+      # @example
+      #
+      #   # bad
+      #   sig { returns(Foo) }
+      #   def foo
+      #     @foo = T.let(@foo, T.nilable(Foo))
+      #     @foo ||= Foo.new
+      #   end
+      #
+      #   # good
+      #   sig { returns(Foo) }
+      #   def foo
+      #     @foo ||= T.let(Foo.new, T.nilable(Foo))
+      #   end
+      #
+      # TODO: disable this cop when the Sorbet version is older than `0.5.10210`.
+      class ObsoleteStrictMemoization < RuboCop::Cop::Cop
+        include RuboCop::Cop::MatchRange
+        include RuboCop::Cop::Alignment
+
+        MESSAGE = "This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. " \
+          "See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization."
+
+        # @!method legacy_memoization_pattern?(node)
+        def_node_matcher :legacy_memoization_pattern?, <<~PATTERN
+          (begin
+            (ivasgn $_ivar                                # @_ivar = ...
+              (send                                       # T.let(_ivar, T.nilable(_ivar_type))
+                (const nil? :T) :let
+                (ivar _ivar)
+                (send                                     # T.nilable(_ivar_type)
+                  (const nil? :T) :nilable $_ivar_type)))
+            (or-asgn                                      # @_ivar ||= _initialization_expr
+              (ivasgn _ivar)
+              $_initialization_expr))
+        PATTERN
+
+        def on_begin(node)
+          return unless legacy_memoization_pattern?(node)
+
+          add_offense(node, message: MESSAGE)
+        end
+
+        def autocorrect(node)
+          ->(corrector) {
+            expression = legacy_memoization_pattern?(node)
+            ivar, ivar_type, initialization_expr = expression
+
+            base_indent = offset(node)
+
+            is_multiline_init_expr = initialization_expr.line_count != 1
+
+            correction = if is_multiline_init_expr
+              render_multi_line_correction(ivar, ivar_type, initialization_expr, base_indent)
+            else
+              single_line_correction =
+                "#{ivar} ||= T.let(#{initialization_expr.source}, T.nilable(#{ivar_type.source}))"
+
+              if (base_indent.length + single_line_correction.length) <= max_line_length
+                single_line_correction
+              else # The single-line correction was too long. Re-render it as a multi-line correction.
+                render_multi_line_correction(ivar, ivar_type, initialization_expr, base_indent)
+              end
+            end
+
+            corrector.replace(node, correction)
+          }
+        end
+
+        private
+
+        def single_indent
+          @single_indent ||= case config.for_cop("Layout/IndentationStyle")["EnforcedStyle"]
+          when "tabs" then "\t"
+          when "spaces", nil then " " * configured_indentation_width
+          end
+        end
+
+        def render_multi_line_correction(ivar, ivar_type, initialization_expr, base_indent)
+          indent = single_indent
+          initialization_expr_source = initialization_expr.source.lines.map { |line| indent + line }.join
+
+          <<~RUBY.chomp
+            #{ivar} ||= T.let(
+            #{base_indent}#{initialization_expr_source},
+            #{base_indent}#{indent}T.nilable(#{ivar_type.source}),
+            #{base_indent})
+          RUBY
+        end
+      end
+    end
+  end
+end
--- a/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
+++ b/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
@@ -20,6 +20,14 @@ module RuboCop
       #     @foo ||= Foo.new
       #   end
       #
+      #   # bad
+      #   sig { returns(Foo) }
+      #   def foo
+      #     # This would have been a mistake, causing the memoized value to be discarded and recomputed on every call.
+      #     @foo = T.let(nil, T.nilable(Foo))
+      #     @foo ||= Foo.new
+      #   end
+      #
       #   # good
       #   sig { returns(Foo) }
       #   def foo
@@ -46,7 +54,7 @@ module RuboCop
             $(ivasgn $_ivar                                           # First line: @_ivar = ...
               (send                                                   # T.let(_ivar, T.nilable(_ivar_type))
                 $(const {nil? cbase} :T) :let
-                (ivar _ivar)
+                {(ivar _ivar) nil}
                 (send (const {nil? cbase} :T) :nilable $_ivar_type))) # T.nilable(_ivar_type)
             $(or-asgn (ivasgn _ivar) $_initialization_expr))          # Second line: @_ivar ||= _initialization_expr
         PATTERN
--- a/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
+++ b/lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
@@ -26,11 +26,13 @@ module RuboCop
       #     @foo ||= T.let(Foo.new, T.nilable(Foo))
       #   end
       #
-      # TODO: disable this cop when the Sorbet version is older than `0.5.10210`.
       class ObsoleteStrictMemoization < RuboCop::Cop::Base
         include RuboCop::Cop::MatchRange
         include RuboCop::Cop::Alignment
         extend AutoCorrector
+
+        include TargetSorbetVersion
+        minimum_target_sorbet_static_version "0.5.10210"
 
         MESSAGE = "This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. " \
           "See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization."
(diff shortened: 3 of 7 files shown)


lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,123 @@
...
+      #     @foo ||= Foo.new
+      #   end
+      #
+      #   # good
+      #   sig { returns(Foo) }
+      #   def foo
+      #     @foo ||= T.let(Foo.new, T.nilable(Foo))
+      #   end
+      #
+      # TODO: disable this cop when the Sorbet version is older than `0.5.10210.20220725105033-6e80c502e`.
```
amomchilov, line 29: How do you do that? Is there something like a `support_target_sorbet_runtime_version?`

https://github.com/rubocop/rubocop/blob/172cc752c39a15439730adb5d8a75058ac18ecf4/lib/rubocop/cop/team.rb#L159-L180

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,123 @@
...
+      # TODO: find the right way to access the line length limit, indentation style, and indentation width.
```
amomchilov, line 30: `RuboCop::Cop::Alignment` looks promising, but I can't find any good docs on it

https://rubydoc.info/gems/rubocop/1.51.0/RuboCop/Cop/Alignment

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,123 @@
...
+
+            when "spaces", nil
+              space_width = ident_style_config["IndentationWidth"] || config.for_cop("Layout/IndentationWidth")["Width"] || 2
+              " " * space_width
+            end
+          end
+        end
+
+        def max_line_length
+          @max_line_length ||= config.for_cop("Layout/LineLength").fetch("Max", 120)
```
amomchilov, line 103: Likewise, this hard-coded `120` is a bit bleh

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+              )
+            end
+          RUBY
+
+          expect(autocorrect_source(source)).to(eq(corrected_source))
+        end
+      end
+    end
+
+    context "when its not the first line in a method", pending: "Not implemented yet" do
```
amomchilov, line 174: I'm not sure how to support this. I can start my node matcher with `...` to metch against `some`/`other`/`code`, but then the replacement will completely delete those lines.

If I do match against them, I'd need to tweak the replacement to skip over them and only replace the correct portion.

More generally, what's the right way to match "any two consecutive lines, anywhere" (ignoring whitespace in between)?

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,129 @@
...
+          @single_indent ||= begin
+            ident_style_config = config.for_cop("Layout/IndentationStyle")
+
+            case ident_style_config["EnforcedStyle"]
+            when "tabs"
+              "\t"
+
+            when "spaces", nil
+              space_width = ident_style_config["IndentationWidth"] ||
+                config.for_cop("Layout/IndentationWidth")["Width"] || 2
```
amomchilov, line 95: This `||` 2 is a bit of a hack.

Is there an API for accessing the config of another cop, and having it automatically apply its own efault value (if there's no explicitly configured value)?

Barring that, is there a way to access a default value for another cop's config?

manual/cops_sorbet.md
```diff
@@ -509,6 +509,38 @@ sig { params(b: String, a: Integer).void }
 def foo(b:, a: 1); end
 ```
 
+## Sorbet/ObsoleteStrictMemoization
+
+Enabled by default | Safe | Supports autocorrection | VersionAdded | VersionChanged
+--- | --- | --- | --- | ---
+Enabled | Yes | No | - | -
```
amomchilov, line 516: Why does `bundle exec rake generate_cops_documentation` think it doesn't support autocorrection? It implements `#autocorrect`.

amomchilov: Ah it happened when I switched from `< RuboCop::Cop::Cop` to `< RuboCop::Cop::Base`, which also broke the (previously passing) tests

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,126 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # This cop checks the obsolete pattern for initializing instance variables that was required for older Sorbet
```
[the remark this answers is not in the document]
amomchilov: Does `for the` make the most sense?

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,126 @@
...
+      #     @foo ||= T.let(Foo.new, T.nilable(Foo))
+      #   end
+      #
+      # TODO: disable this cop when the Sorbet version is older than `0.5.10210.20220725105033-6e80c502e`.
+      # TODO: find the right way to access the line length limit, indentation style, and indentation width.
+      class ObsoleteStrictMemoization < RuboCop::Cop::Base
+        include RuboCop::Cop::MatchRange
+        extend AutoCorrector
+
+        MESSAGE = "This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. " \
```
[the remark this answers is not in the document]
amomchilov: It really was a workaround though, wasn't it?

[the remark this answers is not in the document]
andyw8: It might depend on your perspective, "approach" seems more neutral. No big deal though.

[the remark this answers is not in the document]
amomchilov: Jez himself called it a "workaround": https://stackoverflow.com/a/56694511/3141234

(workaround is not quite as neutral, but it's not "hack" either)

lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
```diff
@@ -0,0 +1,49 @@
...
+
+          self.class.support_target_sorbet_static_version?(target_sorbet_static_version_from_bundler_lock_file)
+        end
+
+        def target_sorbet_static_version_from_bundler_lock_file
+          @target_sorbet_static_version_from_bundler_lock_file ||= read_sorbet_static_version_from_bundler_lock_file
+        end
+
+        # Adapted from https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/config.rb#L293-L308
+        def read_sorbet_static_version_from_bundler_lock_file
```
amomchilov, line 33: This is all pretty yucky. It'd be nice to have a native API to support this

https://github.com/rubocop/rubocop/issues/11920

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
amomchilov: That's an interesting idea!

I'll ask on the Rubocop project, and see if there's a "blessed" way to express this, otherwise that seems like a promising fallback!

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,132 @@
...
+      #     @foo ||= T.let(Foo.new, T.nilable(Foo))
+      #   end
+      #
+      # TODO: find the right way to access the line length limit, indentation style, and indentation width.
+      class ObsoleteStrictMemoization < RuboCop::Cop::Base
+        include RuboCop::Cop::MatchRange
+        extend AutoCorrector
+
+        include TargetSorbetVersion
+        minimum_target_sorbet_static_version "0.5.10210"
```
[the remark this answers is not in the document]
amomchilov: > Should this be a constant?

I don't think so, it would just add indirection without any novel information, e.g.

```suggestion
        MINIMUM_SUPPORTED_VERSION = "0.5.10210"
        minimum_target_sorbet_static_version MINIMUM_SUPPORTED_VERSION
```

or did you mean something else?

> If we expect it to be modified

We don't, this is a quality of the cop (and Sorbet's support for the syntax it corrects to), which wouldn't be user configurable

lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
```diff
@@ -0,0 +1,49 @@
...
+            @minimum_target_sorbet_static_version = Gem::Version.new(version)
+          end
+
+          def support_target_sorbet_static_version?(version)
+            @minimum_target_sorbet_static_version <= Gem::Version.new(version)
+          end
+        end
+
+        def enabled_for_sorbet_static_version?
+          return false unless target_sorbet_static_version_from_bundler_lock_file
```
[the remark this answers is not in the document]
amomchilov: I could, but I figured it's memoized anyway

Should I?

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,132 @@
...
+        MESSAGE = "This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. " \
+          "See https://github.com/sorbet/sorbet/issues/995."
+
+        # @!method legacy_memoization_pattern?(node)
+        def_node_matcher :legacy_memoization_pattern?, <<~PATTERN
+          (begin
+            (ivasgn $_ivar                                # @_ivar = ...
+              (send                                       # T.let(_ivar, T.nilable(_ivar_type))
+                (const nil? :T) :let
+                (ivar _ivar)
```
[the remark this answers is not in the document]
amomchilov: Interesting idea... do you mean this?

```ruby
sig { returns(Foo) }
def foo
  @foo = T.let(nil, T.nilable(Foo))
  #            ^^^
  @foo ||= Foo.new
end
```

That's more of a misapplication of the pattern... seems important to check, since it would disable the memoization behaviour entirely haha

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -0,0 +1,211 @@
...
+require "spec_helper"
+
+RSpec.describe(RuboCop::Cop::Sorbet::ObsoleteStrictMemoization, :config) do
+  before(:each) do
+    allow(cop).to(receive(:target_sorbet_static_version_from_bundler_lock_file).and_return("0.5.10210"))
+  end
+
+  it "the new memoization pattern doesn't register any offense" do
+    expect_no_offenses(<<~RUBY)
+      sig { returns(Foo) }
```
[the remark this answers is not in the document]
amomchilov: Nope! I'll remove them

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,132 @@
...
+      # TODO: find the right way to access the line length limit, indentation style, and indentation width.
+      class ObsoleteStrictMemoization < RuboCop::Cop::Base
+        include RuboCop::Cop::MatchRange
+        extend AutoCorrector
+
+        include TargetSorbetVersion
+        minimum_target_sorbet_static_version "0.5.10210"
+
+        MESSAGE = "This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. " \
+          "See https://github.com/sorbet/sorbet/issues/995."
```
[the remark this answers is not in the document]
amomchilov: Oh fantastic! didn't know we had docs for this

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,132 @@
...
+            (or-asgn                                      # @_ivar ||= _initialization_expr
+              (ivasgn _ivar)
+              $_initialization_expr))
+        PATTERN
+
+        def on_begin(node)
+          return unless legacy_memoization_pattern?(node)
+
+          add_offense(node, message: MESSAGE) do |corrector|
+            expression = legacy_memoization_pattern?(node)
```
dvandersluis, line 58: You probably want to avoid calling the matcher multiple times here.

lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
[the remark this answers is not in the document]
Morriar: Even if the returned value is memoized, you still have to pay for two method calls when you only need one.

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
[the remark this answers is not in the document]
Morriar: No, I mean to not use `minimum_target_sorbet_static_version` at all and just a constant:

```rb
MINIMUM_TARGET_SORBET_STATIC_VERSION = Gem::Version.new(version)
```

And then in the helper method:

```rb
def support_target_sorbet_static_version?(version)
  MINIMUM_TARGET_SORBET_STATIC_VERSION<= Gem::Version.new(version)
end
```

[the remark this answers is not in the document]
amomchilov: Ah, I see what you mean.

I wanted it to be a declarative style, similar to `minimum_target_rails_version` https://github.com/search?q=repo%3Arubocop%2Frubocop-rails+minimum_target_rails_version&type=code

See https://github.com/rubocop/rubocop/issues/11920 for context

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+
+        autocorrected_source = autocorrect_source(<<~RUBY)
+          def foo
+            @foo = T.let(@foo, T.nilable(SomeReallyLongTypeName______________________________________))
+            @foo ||= some_really_long_initialization_expression______________________________________
+          end
+        RUBY
+
+        longest_line = autocorrected_source.lines.max_by(&:length)
+        expect(longest_line.length).to(be <= 120) # FIXME: Unhardcode this 120
```
sambostock, line 81: I assume 120 is meant to match the `Layout/LineLength` `Max`? I think the "correct" way to do this would be to override the config in this test to specify that value explicitly.

e.g. https://github.com/rubocop/rubocop/blob/d69fcd0964a441e085a199deb0f55d2e87731a0d/spec/rubocop/cop/style/safe_navigation_spec.rb#L958-L965
e.g. https://github.com/rubocop/rubocop/blob/d69fcd0964a441e085a199deb0f55d2e87731a0d/spec/rubocop/cop/layout/redundant_line_break_spec.rb#L3-L9

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
```diff
@@ -0,0 +1,123 @@
...
+            $(or-asgn                                     # Second line: @_ivar ||= _initialization_expr
+              (ivasgn _ivar)
+              $_initialization_expr))
+        PATTERN
+
+        def on_begin(node)
+          expression = legacy_memoization_pattern?(node)
+          return unless expression
+
+          first_assignment_node, ivar, ivar_type, second_conditional_assignment_node, initialization_expr = expression
```
sambostock, line 60: I tend to express this with a block rather than a guard clause.

```ruby
legacy_memoization_pattern?(node) do |first_assignment_node, ivar, ivar_type, second_conditional_assignment_node, initialization_expr|
  # ...
end
```

Although this is a bunch of captures. Either way works.

sambostock:
I took a stab at simplifying things [on this branch](https://github.com/Shopify/rubocop-sorbet/compare/obsolete-memoization-cop...obsolete-memoization-cop-alt), if you're interested.

The indentation and correction logic is simplified (relying on other cops to handle things like tabs vs spaces, and replacing the entire complete range, rather than multi-phase replacement), while still passing the existing specs (and adding coverage for the `::T` edge case).

amomchilov:
@sambostock That looks great, I pushed your commit into this branch/PR.

I think this is ready to review and merge!

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
sambostock: I think this might have gotten lost among the previous reviews.

KaanOzkan (approved):
LGTM, there are a few open comments from @Morriar so let's wait for his review please 🙂

config/default.yml
```diff
@@ -151,6 +151,18 @@ Sorbet/KeywordArgumentOrdering:
...
   VersionAdded: 0.2.0
 
+Sorbet/ObsoleteStrictMemoization:
+  Description: >-
+    This cop checks for the obsolete pattern for initializing instance variables that was required for older Sorbet
+    versions in `#typed: strict` files.
+
+    It's no longer required, as of Sorbet 0.5.10210
+    See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization
+  Enabled: pending
```
KaanOzkan, line 161: We haven't been releasing major versions yet and from what I'm reading `pending` will stay disabled unless there is a major bump or `NewCops` setting is enabled. Given this should we enable it by default?

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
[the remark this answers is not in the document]
Morriar: I counted very approximatively how many of these cases we have in core (`git grep -A 1 "@.* = T.let(nil," | grep -B 1 ".*\.new"`) and it seems it's only a handful, maybe we can fix or disregard them 👍

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
amomchilov: Indeed it was, thanks for the nudge and the pointer. This approach worked great!

lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+
+        # Adapted from https://github.com/rubocop/rubocop/blob/1181d4ebad5f71c586f9514d9c341cdfffc1957d/lib/rubocop/config.rb#L293-L308
+        def read_sorbet_static_version_from_bundler_lock_file
+          lock_file_path = config.bundler_lock_file_path
+
+          return nil unless lock_file_path
+
+          File.foreach(lock_file_path) do |line|
+            # If Sorbet (or one of its frameworks) is in Gemfile.lock or gems.lock, there should be
```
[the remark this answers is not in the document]
amomchilov: This comes from https://github.com/rubocop/rubocop/blob/3bf6d4132736ac228efdaf0e85901acbaa968ce0/lib/rubocop/config.rb#L293-L308

I'd like to keep it the same until hopefully one day it can be deleted (replaced with some Rubocop API that let's you specify min version reqs).

lib/rubocop/cop/sorbet/obsolete_strict_memoization.rb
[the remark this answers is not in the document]
amomchilov: Implemented!

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -221,4 +221,25 @@ def foo
       end
     end
   end
+
+  describe "a mistaken variant of the obsolete memoization pattern" do
+    it "registers an offence and autocorrects" do
+      # This variant would have been a mistake, which woudl have caused the memoized value to be discarded and recomputed
```
sambostock, line 227:
```suggestion
      # This variant would have been a mistake, which would have caused the memoized value to be discarded and recomputed
```

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -221,4 +221,25 @@ def foo
...
+      # on every call. We can fix it up into the working version.
+
+      expect_offense(<<~RUBY)
+        def foo
+          @foo = T.let(nil, T.nilable(Foo))
+          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This two-stage workaround for memoization in `#typed: strict` files is no longer necessary. See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization.
```
sambostock, line 233: Should we change the error message? This isn't just changing the syntax, it's fixing a bug.

lib/rubocop/cop/sorbet/mixin/target_sorbet_version.rb
```diff
@@ -29,22 +29,18 @@ def enabled_for_sorbet_static_version?
         end
 
         def target_sorbet_static_version_from_bundler_lock_file
-          @target_sorbet_static_version_from_bundler_lock_file ||= read_sorbet_static_version_from_bundler_lock_file
+          if defined?(@target_sorbet_static_version_from_bundler_lock_file)
```
Morriar, line 32: Why do we need this change?

Morriar: Okay, it can be nil and we don't want to reread it every time 👍

Morriar (approved):

spec/rubocop/cop/sorbet/obsolete_strict_memoization_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+        end
+      RUBY
+
+      expect_correction(<<~RUBY)
+        def foo
+          @foo ||= T.let(Foo.new, T.nilable(Foo))
+        end
+      RUBY
+    end
+  end
```
amomchilov, line 284: This correction is unsafe, so I'm extracting it to a new `BuggyObsoleteMemoizationCopy` that's correctly marked `SafeAutoCorrect: false`:

https://github.com/Shopify/rubocop-sorbet/pull/175

PR #164, opened 2023-07-12
Fix signature order cop when we have an incomplete call

vinistock:
While the developer is typing, we might have no syntax errors, but an incomplete signature still (e.g.: `sig { params(a: Integer).voi }`).

This causes the cop to break, which is surfaced by the Ruby LSP basically every single time someone is typing a signature.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
@@ -36,6 +36,10 @@ module RuboCop
           calls = call_chain(node.children[2]).map(&:method_name)
           return unless calls.any?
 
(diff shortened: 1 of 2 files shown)


Morriar (approved):

vinistock:
Yeah, it's worth investigating techniques to find these. The only issue with going character by character is that it might not cover all possible scenarios. For example

```ruby
# This error
params().

# Is different than this
params()v

# Which is also different than
params.v
```

Depending on autocompletion and auto-closing brackets, there might be states that aren't captured by going char by char. At the same time, it produces a lot of redundant tests since

```ruby
params().v
params().vo
params().voi
```
are all the same test case for this particular scenario. It might be worth connecting with the YARP team and asking how they are coming up with partially complete code examples.

Issue #163, opened 2023-07-09
Forbid defining implicit conversion methods (to_ary, to_hash, to_int, to_str)


dduugg:
Ruby supports a few methods that allow implicit type [conversions](https://docs.ruby-lang.org/en/3.2/implicit_conversion_rdoc.html).

Since Sorbet is a nominal (not [structural](https://sorbet.org/docs/faq#can-i-use-sorbet-for-duck-typed-code)) type system, implicit conversion is currently [unsupported](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Pathname%0A%20%20extend%20T%3A%3ASig%0A%20%20sig%20%7B%20returns%28String%29%20%7D%0A%20%20def%20to_str%0A%20%20%20%20to_s%0A%20%20end%0Aend%0A%0A''%20%2B%20Pathname%28'%2F'%29%20%23%20%3D%3E%20%22%2F%22).

I would like to write a cop that adds violations when implicit conversion methods are defined. Would the maintainers be amenable to this?

Morriar:
Yeah it seems like a good idea to have such a cop 👍 

Depending on the amount of violations it creates, we may not enable it by default though.

PR #165, opened 2023-07-13
Add new `Sorbet/ImplicitConversionMethod` cop

dduugg:
Resolves https://github.com/Shopify/rubocop-sorbet/issues/163

Adds a cop to forbid implicit conversion methods. For conciseness, the cop name is consistent with rubocop core's style of naming the thing that is disallowed, (e.g. `DuplicateMethod`), rather than the more verbose `Forbid*` style (e.g. `ForbidTUntyped`).


Rakefile
```diff
@@ -18,7 +18,7 @@ task :new_cop, [:cop] do |_task, args|
   require "rubocop"
 
   cop_name = args.fetch(:cop) do
-    warn("usage: bundle exec rake new_cop[Department/Name]")
+    warn('usage: bundle exec rake "new_cop[Department/Name]"')
```
Morriar, line 21: Just in case we need to revert the current change, do you mind moving this to a separate PR please 🙏

lib/rubocop/cop/sorbet/implicit_conversion_method.rb
```diff
@@ -0,0 +1,55 @@
...
+      #   def self.to_str(x); end
+      #
+      #   # bad
+      #   alias to_str to_s
+      #
+      # @see https://docs.ruby-lang.org/en/master/implicit_conversion_rdoc.html
+      # @note Since the arity of aliased methods is not checked, false positives may result.
+      class ImplicitConversionMethod < RuboCop::Cop::Base
+        IMPLICIT_CONVERSION_METHODS = [:to_ary, :to_int, :to_hash, :to_str].freeze
+        MSG = "Avoid implicit conversion methods, as Sorbet does not support them."
```
Morriar, line 33: Should we suggest an alternative?

dduugg:
@Morriar I believe I've addressed the initial PR comments, PTAL

lib/rubocop/cop/sorbet/implicit_conversion_method.rb
```diff
@@ -0,0 +1,56 @@
...
+        def on_def(node)
+          return unless IMPLICIT_CONVERSION_METHODS.include?(node.method_name)
+          return unless node.arguments.empty?
+
+          add_offense(node)
+        end
+        alias_method :on_defs, :on_def
+
+        def on_send(node)
+          add_offense(node.first_argument) if IMPLICIT_CONVERSION_METHODS.include?(node.first_argument.value)
```
Morriar, line 51: This is going to add an offense for call as well right? Should we add this to the documentation and also add a test for it?

lib/rubocop/cop/sorbet/implicit_conversion_method.rb
```diff
@@ -0,0 +1,56 @@
...
+          add_offense(new_id) if IMPLICIT_CONVERSION_METHODS.include?(new_id.value)
+        end
+
+        def on_def(node)
+          return unless IMPLICIT_CONVERSION_METHODS.include?(node.method_name)
+          return unless node.arguments.empty?
+
+          add_offense(node)
+        end
+        alias_method :on_defs, :on_def
```
Morriar, line 48: Do you mind adding a test for this. I see it's already properly documented.

Morriar (commented):
Thanks a lot!

Just two details and it should be good to go 👍

PR #166, opened 2023-07-17
Suggest zsh-compatible rake syntax

dduugg:
As requested by @Morriar in https://github.com/Shopify/rubocop-sorbet/pull/165#discussion_r1262871469


dduugg:
> Should we also update the command in the README while at it?
> 
> https://github.com/Shopify/rubocop-sorbet/blob/0a4c5ee865bfb60a1fe3eda5958dba193c0fbb1d/README.md?plain=1#L101

👍 PTAL

Issue #169, opened 2023-08-04
ValidSigil parsing behaves differently than sorbet sigil parsing


jakebrady5:
### Bug Description

The `ValidSigil` cop considers any valid sigil value followed by any non-word non-whitespace character to be valid. Sorbet does not. This conflict causes the `ValidSigil` cop to be overly permissive of functionally invalid sigil values.

### Practical Example

A sample file began with this sigil:

`# typed: false # rubocop:todo Sorbet/StrictSigil`

A similar regex [bug](https://github.com/Shopify/spoom/issues/427) in `spoom` caused spoom to erroneously bump the sigil to:

`# typed: true# rubocop:todo Sorbet/StrictSigil`

This new `true#` sigil is functionally ignored by Sorbet, but does not cause a rubocop error despite being an invalid sigil value.

PR #170, opened 2023-08-04
match SIGIL_REGEX behavior to sorbet sigil parsing

jakebrady5:
closes https://github.com/Shopify/rubocop-sorbet/issues/169

This change makes the `ValidSigil` cop's sigil parsing match behavior more closely to the sorbet [implementation](https://github.com/sorbet/sorbet/blob/10541602dda21d749ffeb5ef6cb8401b26e4365f/core/Files.cc#L24-L96).

### DIFF
--- a/lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
+++ b/lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
@@ -51,7 +51,7 @@ module RuboCop
         protected
 
         STRICTNESS_LEVELS = ["ignore", "false", "true", "strict", "strong"]
-        SIGIL_REGEX = /^\s*#\s+typed:(?:\s+([\w]+))?/
+        SIGIL_REGEX = /^\s*#\s+typed:(?:\s+([\S]+))?/
 
         # extraction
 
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
```diff
@@ -51,7 +51,7 @@ def autocorrect(_node)
         protected
 
         STRICTNESS_LEVELS = ["ignore", "false", "true", "strict", "strong"]
-        SIGIL_REGEX = /^\s*#\s+typed:(?:\s+([\w]+))?/
+        SIGIL_REGEX = /^\s*#\s+typed:(?:\s+([\S]+))?/
```
paracycle, line 54: While we are at it, can we also make these `\s` instances into `[[:blank:]]` since we don't actually want to match newlines, etc, just blank space:
```suggestion
        SIGIL_REGEX = /^[[:blank:]]*#[[:blank:]]+typed:(?:[[:blank:]]+([\S]+))?/
```

jakebrady5:
Thanks @vinistock! Minor version bump and release would be 🥇 if possible. Happy to bump version here if you'd prefer.

PR #172, opened 2023-08-14
Update RedundantExtendTSig to trim whole line

bitwise-aiden:
### Description 

When running this cop it will leave whatever whitespace occurred at the start of the line, which means that if run in isolation may cause other violations. 

This approach is in line with what occurs in [ForbidExtendTSigHelpersInShims](https://github.com/Shopify/rubocop-sorbet/blob/main/lib/rubocop/cop/sorbet/rbi/forbid_extend_t_sig_helpers_in_shims.rb#L40), replacing the whole line and removing any new line. 

The thing to consider is that it will also remove any comment on the line, but this feels like a fine tradeoff since it would likely be to do with the extend in the first place and would need to be cleaned up.

### DIFF
--- a/lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
+++ b/lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
@@ -26,6 +26,7 @@ module RuboCop
       #   end
       #
       class RedundantExtendTSig < RuboCop::Cop::Base
+        include RangeHelp
         extend AutoCorrector
 
         MSG = "Do not redundantly `extend T::Sig` when it is already included in all modules."
@@ -40,7 +41,7 @@ module RuboCop
           return unless extend_t_sig?(node)
 
           add_offense(node) do |corrector|
-            corrector.remove(node)
+            corrector.remove(range_by_whole_lines(node.source_range, include_final_newline: true))
           end
         end
       end
--- a/spec/rubocop/cop/sorbet/redundant_extend_t_sig_spec.rb
+++ b/spec/rubocop/cop/sorbet/redundant_extend_t_sig_spec.rb
@@ -14,7 +14,6 @@ RSpec.describe(RuboCop::Cop::Sorbet::RedundantExtendTSig, :config) do
 
       expect_correction(<<~RUBY)
         #{header}
-          #{trailing_whitespace}
         end
       RUBY
     end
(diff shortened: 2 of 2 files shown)


egiurleo (approved):

lib/rubocop/cop/sorbet/redundant_extend_t_sig.rb
```diff
@@ -40,7 +41,7 @@ def on_send(node)
           return unless extend_t_sig?(node)
 
           add_offense(node) do |corrector|
-            corrector.remove(node)
+            corrector.remove(range_by_whole_lines(node.source_range, include_final_newline: true))
```
paracycle, line 44: What if we did `range_with_surrounding_space(range: node.source_range)` instead?

bitwise-aiden: That's a lot cleaner. I'm guessing this will also take out additional newlines before and after?

bitwise-aiden: Ahh, that could be why 
image

bitwise-aiden: [Looking at the docs](https://www.rubydoc.info/gems/rubocop/0.50.0/RuboCop/Cop/Util.range_with_surrounding_space), I wonder how many cases would be solved with:
```
range_with_surrounding_space(range: node.source_range, side: :right) 
```

bitwise-aiden: Ahh, not quite! This one is wrong:
```ruby
class SomeClass 
  extend T::Sig
end 

class SomeClass
  end
```

The indentation is kept because it doesn't go left. Though that becomes tough to manage since if it was anything other than `end` we would be fine with it.

Maybe it is best to keep it how it was and rely on the other cops to ensure the lines are cleaned up.

paracycle: Interesting, what if we did: `range_with_surrounding_space(range: node.source_range, side: :right)` and joined it with `range_with_surrounding_space(range: node.source_range, side: :left, newlines: false)`. So:
```ruby
add_range(
  range_with_surrounding_space(range: node.source_range, side: :left, newlines: false),
  range_with_surrounding_space(range: node.source_range, side: :right)
)
```

bitwise-aiden: Thanks for the inquiry, it's good to know what is going on at a deeper level.

PR #175, opened 2023-08-31
Extract new `BuggyObsoleteStrictMemoization` cop

amomchilov:
The `Sorbet/ObsoleteStrictMemoization` cop introduced in #162 was initially meant to detect this kind of pattern:

```ruby
sig { returns(Foo) }
def foo
  @foo = T.let(@foo, T.nilable(Foo))
  @foo ||= Foo.new
end
```

@Morriar had [the fantastic idea](https://github.com/Shopify/rubocop-sorbet/pull/162#discussion_r1220133334) to also detect a mistaken variant:

```ruby
sig { returns(Foo) }
def foo
  # This would have been a mistake, causing the memoized value to be discarded and recomputed on every call.
  @foo = T.let(nil, T.nilable(Foo))
  #            !!!
  @foo ||= Foo.new
end
```

This was a great idea and found several such bugs in our program, but wasn't implemented quite right.

Auto-correcting this is dangerous. If the computation being memoized (`Foo.new`, in this case) had side effects, calling it only once (instead of once on every call to `foo`) can be observed, and might be a breaking change. This is problematic because `Sorbet/ObsoleteStrictMemoization` is marked `Safe: true` and `SafeAutoCorrect: true`.

This PR splits off this behaviour into a new `BuggyObsoleteStrictMemoization` cop. The output of this cop is the correct (but still obsolete) memoization pattern. From there, `ObsoleteStrictMemoization` can be applied to bring it to the modern standard. This cop is marked `Safe: true, SafeAutoCorrect: false`.

### DIFF
--- a/lib/rubocop/cop/sorbet/buggy_obsolete_strict_memoization.rb
+++ b/lib/rubocop/cop/sorbet/buggy_obsolete_strict_memoization.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Checks for the obsolete pattern for initializing instance variables that was required for older Sorbet
+      # versions in `#typed: strict` files.
+      #
+      # It's no longer required, as of Sorbet 0.5.10210
+      # See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization
+      #
+      # @example
+      #
+      #   # bad
+      #   sig { returns(Foo) }
+      #   def foo
+      #     @foo = T.let(@foo, T.nilable(Foo))
+      #     @foo ||= Foo.new
+      #   end
+      #
+      #   # bad
+      #   sig { returns(Foo) }
+      #   def foo
+      #     # This would have been a mistake, causing the memoized value to be discarded and recomputed on every call.
+      #     @foo = T.let(nil, T.nilable(Foo))
+      #     @foo ||= Foo.new
+      #   end
+      #
+      #   # good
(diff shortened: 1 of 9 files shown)


spec/rubocop/cop/sorbet/buggy_obsolete_strict_memoization_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+    end
+
+    it "registers an offence and autocorrects" do
+      # This variant would have been a mistake, which would have caused the memoized value to be discarded
+      # and recomputed on every call. We can fix it up into the working version.
+
+      expect_offense(<<~RUBY)
+        def foo
+          @foo = T.let(nil, T.nilable(Foo))
+                       ^^^ This might be a mistaken variant of the two-stage workaround that used to be needed for memoization in `#typed: strict` files. See https://sorbet.org/docs/type-assertions#put-type-assertions-behind-memoization.
```
amomchilov, line 136: Kinda nice, the `^^^` is now only under the `nil` part of this, rather than the whole line.

lib/rubocop/cop/sorbet/buggy_obsolete_strict_memoization.rb
amomchilov, line 1: For an easier reviewing experience, take a look at the commit list.

I made this by first copying over the `ObsoleteStrictMemoization` as-is, then making changes to it in a second commit. YoSo you can just review the diff to go from `ObsoleteStrictMemoization` to `BuggyObsoleteStrictMemoization`

andyw8 (approved):

PR #177, opened 2023-09-11
Add Sorbet/ForbidTStruct cop

bitwise-aiden:
Through performance benchmarks, we've found that `T::Struct` is up to 5x slower than it's PORO counterpart. Along with being a catalyst for potential code smells, we've decided to create this cop to forbid the future usage of `T::Struct` in new code paths.


lib/rubocop/cop/sorbet/forbid_t_struct.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Disallows using `T::Struct`.
+      #
+      # `T::Struct` has runtime performance implications that can lead to around 5x slower code over the equivilant
```
andyw8, line 10:
```suggestion
      # `T::Struct` has runtime performance implications that can lead to around 5x slower code over the equivalent
```

andyw8: Maybe better to remove the 5x part since that could change over time if Sorbet improves?

PR #181, opened 2023-10-06
Forbid usage of type aliased shapes

KaanOzkan:


spec/rubocop/cop/sorbet/forbid_type_aliased_shapes_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+require "spec_helper"
+
+RSpec.describe(RuboCop::Cop::Sorbet::ForbidTypeAliasedShapes, :config) do
+  it("allows defining type aliases that don't contain shapes") do
+    expect_no_offenses(<<~RUBY)
+      Foo = T.type_alias { Integer }
+    RUBY
+  end
+
+  it("disallows defining type aliases that contain shapes") do
```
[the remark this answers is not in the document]
Morriar: I don't think this approach will work to catch a shape deeply nested inside the type alias. Consider this: `T.nilable({ foo: Integer })`.

I think we need to descend in the type alias block recursively.

sambostock:
I think it makes sense to have it be a single cop, and make it configurable to allow some behaviours, if there's a use case for it.

lib/rubocop/cop/sorbet/forbid_type_aliased_shapes.rb
```diff
@@ -0,0 +1,66 @@
...
+      #
+      #    sig { params(foo: Integer).void }
+      #    def initialize(foo)
+      #      @foo = foo
+      #    end
+      #  end
+      class ForbidTypeAliasedShapes < RuboCop::Cop::Base
+        MSG = "Type aliases shouldn't contain shapes because of significant performance overhead"
+
+        # @!method type_alias?(node)
```
Morriar, line 27: Is this actually used?

lib/rubocop/cop/sorbet/forbid_type_aliased_shapes.rb
```diff
@@ -0,0 +1,66 @@
...
+        def_node_matcher(:type_alias?, <<-PATTERN)
+          (block
+            (send
+              (const nil? :T) :type_alias)
+            (args)
+            (hash ...)
+          )
+        PATTERN
+
+        # @!method nested_type_alias?(node)
```
Morriar, line 37: Same question here?

lib/rubocop/cop/sorbet/forbid_type_aliased_shapes.rb
KaanOzkan: 😬 Oops

PR #182, opened 2023-10-10
Add `Sorbet/TStructPropertyAttrMacro`

sambostock:
In a `T::Struct`, `attr_*` methods shouldn't be used for properties.

```diff
 class Foo < T::Struct
-  attr_accessor :bar
-  const :bar, ...
+  prop :bar, ...
 end
```


spec/rubocop/cop/sorbet/t_struct_property_attr_macro_spec.rb
```diff
@@ -0,0 +1,297 @@
+# frozen_string_literal: true
+
+RSpec.describe(RuboCop::Cop::Sorbet::TStructPropertyAttrMacro, :config) do
+  context("when using `const` property") do
+    context("and `attr_reader` is used") do
+      context("and the names match") do
+        it("registers an offense") do
+          expect_offense(<<~RUBY)
+            class Foo < T::Struct
```
sambostock, line 9: We should detect this in other classes as well (`T::ImmutableStruct`, and probably others too).

spec/rubocop/cop/sorbet/t_struct_property_attr_macro_spec.rb
```diff
@@ -0,0 +1,297 @@
...
+        RUBY
+      end
+    end
+
+    context("and `attr_writer` is used") do
+      context("and the names match") do
+        it("registers an offense") do
+          expect_offense(<<~RUBY)
+            class Foo < T::Struct
+              attr_writer :bar
```
sambostock, line 63: One case in which this might be legit would be if the writer's `sig` is **more** restrictive:

```ruby
class Foo < T::Struct
  sig { param(number: Integer) }
  attr_writer :number
  
  const :number, Numeric
end
```

_Less_ restrictive would be an error, although would be hard to check for.

This would be pretty unusual though, and arguably a better named method would probably make more sense, so it's probably safe to just assume that if `attr_*` is being used, it's incorrect usage, and let the author of such an exotic usage simply disable the cop.

PR #183, opened 2023-10-12
Use mixin instead of superclass for signature cops

sambostock:
This replaces the `SignatureCop` parent class with a `SignatureHelp` mixin module. Since `SignatureCop` was inheriting from the deprecated `RuboCop::Cop::Cop`, the cops that were inheriting from `SignatureCop` are updated to inherit directly from `Cop` instead.

This approach
- matches what RuboCop does internally,
- avoids the need to override the registry (?), and
- allows granularly migrating from the deprecated `Cop` class to `Base`.

### Before Merging
- [x] Rebase after #186 is merged

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
+++ b/lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-require "rubocop"
 require "stringio"
-require_relative "signature_cop"
 
 module RuboCop
   module Cop
@@ -26,7 +24,9 @@ module RuboCop
       #
       # * `ParameterTypePlaceholder`: placeholders used for parameter types (default: 'T.untyped')
       # * `ReturnTypePlaceholder`: placeholders used for return types (default: 'T.untyped')
-      class EnforceSignatures < SignatureCop
+      class EnforceSignatures < ::RuboCop::Cop::Cop # rubocop:todo InternalAffairs/InheritDeprecatedCopClass
+        include SignatureHelp
+
         def initialize(config = nil, options = nil)
           super(config, options)
           @last_sig_for_scope = {}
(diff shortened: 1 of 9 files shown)


lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
```diff
@@ -93,11 +93,11 @@ def check_node(node)
         end
 
         def param_type_placeholder
-          cop_config["ParameterTypePlaceholder"] || "T.untyped"
+          cop_config["ParameterTypePlaceholder"] || "T.untyped" # rubocop:todo InternalAffairs/UndefinedConfig
```
sambostock, line 96: Due to a bug, it doesn't seem possible to make this cop pass, so I'm disabling it inline.

#184 will disable it entirely until the bug is fixed, at which point we can address all the offenses.

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -12,7 +9,11 @@
 module RuboCop
   module Cop
     module Sorbet
-      class SignatureBuildOrder < SignatureCop
+      #
+      #
+      class SignatureBuildOrder < ::RuboCop::Cop::Cop # rubocop:todo InternalAffairs/InheritDeprecatedCopClass
```
sambostock, line 14: This cop is currently undocumented. Without the empty comments, YARD uses the `rubocop:disable` comment as documentation, which makes it into `manual/cops_sorbet.md`. Adding these explicitly provides empty documentation, which maintains the status quo, as documenting it is out of scope of this PR.

paracycle (approved):

sambostock:
Right, I did intend to do that. 👍 

This PR branches off #186. If we can get that in first, I can do it in a single rebase.

sambostock:
@Morriar I've rebased and removed the `rubocop:disable` comments. 👍

PR #189, opened 2023-10-15
Add `final`, `bind`, & `implementation` to `SignatureBuildOrder`

sambostock:
This teaches `SignatureBuildOrder` to also enforce the ordering of [`final`](https://github.com/sorbet/sorbet/blob/4ef2b632ae1433ed1a544fd568f3481a4fc8f2df/rbi/sorbet/builder.rbi#L9-L10), [`bind`](https://github.com/sorbet/sorbet/blob/4ef2b632ae1433ed1a544fd568f3481a4fc8f2df/rbi/sorbet/builder.rbi#L21-L22), and [`implementation`](https://github.com/sorbet/sorbet/blob/4ef2b632ae1433ed1a544fd568f3481a4fc8f2df/rbi/sorbet/builder.rbi#L12-L13) builder methods.

As per https://github.com/Shopify/rubocop-sorbet/pull/186#discussion_r1358707761

It also includes a refactor which
- **drops the need for `unparser`**, and
- teaches `SignatureBuildOrder` to handle `sig`s which include unknown builder methods, rather than ignore them

### Before Merging

- [x] Rebase after #186 is merged.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
+++ b/lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
@@ -9,12 +9,11 @@ end
 module RuboCop
   module Cop
     module Sorbet
-      # Checks for the correct order of sig builder methods:
-      # - abstract, override, or overridable
-      # - type_parameters
-      # - params
-      # - returns, or void
-      # - soft, checked, or on_failure
+      # Checks for the correct order of `sig` builder methods.
+      #
+      # Options:
+      #
+      # * `Order`: The order in which to enforce the builder methods are called.
       #
       # @example
       #   # bad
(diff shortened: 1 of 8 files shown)


lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -37,13 +37,17 @@ class SignatureBuildOrder < SignatureCop
             :abstract,
             :override,
             :overridable,
+            :final,
```
sambostock, line 40: As per @paracycle in Slack:

> `final` is usually after `abstract`, `override`, `overridable`

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -37,13 +37,17 @@ class SignatureBuildOrder < SignatureCop
...
             :type_parameters,
             :params,
+            :bind,
```
sambostock, line 43: As per @paracycle in Slack:

> `bind` is usually after `params`

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -37,13 +37,17 @@ class SignatureBuildOrder < SignatureCop
...
             :returns,
             :void,
             :soft,
             :checked,
             :on_failure,
+
+            # :implementation, # Where does this go??
```
sambostock, line 50: As per @paracycle in Slack:

> I have no idea what `implementation` is :joy:

sambostock (commented):
We can probably skip specs for these, as the existing specs don't exhaustively cover the builder methods.

The documentation should be updated though.

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
paracycle: Actually, I think I was wrong in that message. First of all `abstract.final` does not make sense and `final.override` makes more sense than `override.final`. `final` has more signal, so probably should come first. WDYT @Morriar?

paracycle:
> What if we took the order from an array in the configuration and we ship with a default (so we can use our preferred order and remove the old sig builders) but it enables people to use their own order and builders?

Yeah, I like this idea...

sambostock:
Okay, I've updated the cop to extract a config, and added the missing methods to that config.

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
```diff
@@ -31,34 +30,22 @@ module Sorbet
...
-          return if calls.any? { |call| ORDER[call].nil? }
-
-          expected_order = calls.sort_by { |call| ORDER[call] }
+          expected_order = calls.sort_by do |call|
+            builder_method_indexes.fetch(call) do
+              # Abort if we don't have a configured order for this call,
+              # likely because the method name is still being typed.
+              return nil
+            end
+          end
```
sambostock, line 48: I refactored this to perform the check during the same loop as generating the sort keys, and tried to make the comment more focused on the impact than the implementation.

Interestingly, having just `return` without `nil` here in a block is an offense.

sambostock:
I've added a test for the config change, and while doing that I realized I could improve the implementation to remove the need for `unparser` and handle unknown builder methods (by just ignoring them, rather than ignoring the entire `sig`).

lib/rubocop/cop/sorbet/signatures/signature_build_order.rb
sambostock: This code has been refactored, so this is no longer relevant.

Morriar (approved):

PR #192, opened 2023-10-27
Disable Style/Empty method for RBIs

andyw8:
As discussed with @Morriar, if `Style/EmptyMethod` ([docs](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyMethod)) is enabled then it will prevent us from breaking up long argument lists in shims.

So instead of writing:

```ruby
  def remote_test(
    long_argument_name_1:,
    long_argument_name_2:,
    long_argument_name_3:
    )
  end
```

You would need to write

```ruby
  def remote_test(long_argument_name_1:, long_argument_name_2:, long_argument_name_3:); end
```

### DIFF
--- a/config/rbi.yml
+++ b/config/rbi.yml
@@ -245,10 +245,6 @@ Style/DefWithParentheses:
 Sorbet/EmptyLineAfterSig:
   Enabled: true
 
-Style/EmptyMethod:
-  Enabled: true
-  EnforcedStyle: compact
-
 Style/FrozenStringLiteralComment:
   Enabled: true
   EnforcedStyle: never


st0012 (approved):

vinistock:
Is it possible to configure this rule based on the argument count? For short definitions like `def foo; end`, it would be nice to maintain the one liners.

vinistock:
In that case, are we okay with the tradeoff we're making? We're favouring more flexibility for long method signatures at the expense of having more people doing this:

```ruby
sig { void }
def foo
end

sig { void }
def bar
end

sig { void }
def baz
end
```

Not trying to block the PR at all, but I wanted to raise this to ensure that we're making this tradeoff consciously and asking ourselves which one hurts the most.

PR #198, opened 2024-02-02
add missing markdown code fence around example

halorrr:
This example was missing the code block fencing around it. So just adding that in.


halorrr:
I have signed the CLA!

PR #199, opened 2024-02-02
Proof of concept: Create a cop to validate gem version annotations in RBI files

egiurleo:
Created as part of hackdays, along with https://github.com/Shopify/rbi/pull/180 and https://github.com/Shopify/tapioca/pull/1585

While adding gem version annotations to RBI files would allow developers to write more comprehensive RBIs for their gems, it has has the potential to create RBI files that are cluttered, disorganized, or incorrect.

This would be the first of a few cops meant to keep versioned RBIs clean and accurate.

This cop checks that every version included in a "@version" annotation fits the format specified by the RBI library.


lib/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations.rb
```diff
@@ -0,0 +1,61 @@
...
+        def valid_version?(version_string)
+          parts = version_string.strip.split(" ")
+          operator, version = parts
+
+          return false if operator.nil? || parts.nil?
+
+          return false unless VALID_OPERATORS.include?(operator)
+
+          begin
+            Gem::Version.new(version)
```
KaanOzkan, line 51: Wdyt of using the underlying method instead? https://github.com/rubygems/rubygems/blob/40784bb244f7471fa938a421e2773ec2c7a45405/lib/rubygems/version.rb#L173C12-L177.

It won't rely on exception throwing but it might be more brittle to future changes. Although I doubt the interface will go through frequent changes.

config/default.yml
```diff
@@ -293,3 +293,15 @@ Sorbet/ValidSigil:
...
   - script/**/*
+
+Sorbet/ValidGemVersionAnnotations:
+  Description: >-
+                  Ensures all gem version annotations in RBI files are correctly formatted per
+                  Ruby's gem version specification guidelines.
+
+                  See the rubygems.org documentation for more information on how to format gem
+                  versions: https://guides.rubygems.org/patterns/#pessimistic-version-constraint
+  Enabled: true
```
Morriar, line 304: I'm not sure it safe to enable by default since shims or generated gem RBIs may have surprising comments.

Don't we intend Tapioca to strip the version tagging when pulling annotations? I think we also talked about the possibility to merge annotations directly into the generated gem RBIs right?

spec/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  it "does not register an offense when comment uses OR version annotations" do
+    expect_no_offenses(<<~RUBY)
+      # @version > 1.3.6
+      # @version <= 4
+    RUBY
+  end
+
+  it "registers an offense when gem version is not formatted correctly" do
```
Morriar, line 29: Let's add a test for an empty `@version` without an operator and one for `@version foo`.

lib/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations.rb
egiurleo: Oh nice, that's a lot simpler! Thank you!

spec/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations_spec.rb
egiurleo: Done!

config/default.yml
```diff
@@ -317,3 +317,15 @@ Sorbet/ForbidComparableTEnum:
   Description: 'Disallows including the `Comparable` module in a `T::Enum`.'
   Enabled: true
   VersionAdded: 0.8.2
+
```
Morriar, line 320: Should this go in the `config/rbi.yml` instead?

lib/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations.rb
```diff
@@ -0,0 +1,63 @@
...
+
+            if gem_versions.empty?
+              message = format(MSG, versions: "empty version")
+              add_offense(comment, message: message)
+              break
+            end
+
+            invalid_versions = gem_versions.select do |version|
+              !valid_version?(version)
+            end
```
Morriar, line 39:
```suggestion
            invalid_versions = gem_versions.reject do |version|
              valid_version?(version)
            end
```

lib/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+        end
+
+        private
+
+        def valid_version?(version_string)
+          parts = version_string.strip.split(" ")
+          operator, version = parts
+
+          return false if operator.nil? || parts.nil?
```
Morriar, line 54:
```suggestion
          return false unless operator && parts
```

Morriar (commented):
Should we make sure the annotation is attached to a class/method/attr node? Or we don't care about this yet?

config/default.yml
egiurleo: Oh I didn't know that existed 🤦🏻

egiurleo:
@Morriar Great question -- how do you generally think about what goes in the same Rubocop rule vs different rules? Like I could imagine that being in a rule called "NoOrphanedVersionAnnotations" or something like that?

lib/rubocop/cop/sorbet/rbi_versioning/gem_version_annotation_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+        end
+
+        private
+
+        def gem_version_annotation?(comment)
+          comment.text.start_with?(VERSION_PREFIX)
+        end
+
+        def gem_versions(comment)
+          comment.text.delete_prefix(VERSION_PREFIX).split(", ")
```
Morriar, line 22: Should we trim the blank spaces for the split string? Should we also use `/, ?/` to split? I'm not sure how strict we want to be in this cop?

Morriar:
@egiurleo I think it would be better in the same cop but we can merge this now and iterate in another PR 👍

Morriar (approved):

lib/rubocop/cop/sorbet/rbi_versioning/valid_gem_version_annotations.rb
```diff
@@ -0,0 +1,62 @@
...
+      #   # good
+      #   # @version = 1
+      #
+      #   # good
+      #   # @version > 1.2.3
+      #
+      #   # good
+      #   # @version <= 4.3-preview
+      #
+      class ValidGemVersionAnnotations < Base
```
andyw8, line 21:
```suggestion
      class ValidGemVersionAnnotations < RuboCop::Cop::Base
```
(https://github.com/Shopify/rubocop-sorbet/pull/246)

lib/rubocop/cop/sorbet/rbi_versioning/gem_version_annotation_helper.rb
egiurleo: No I like this better, it makes everything more flexible.

PR #204, opened 2024-03-19
Reimplement `CallbackConditionalsBinding` cop & specs

sambostock:
This (in three separate commits, for ease of review)

- **Improves the `CallbackConditionalsBinding` specs** 

  Specs are consolidated by using `expect_correction` to test offenses and correction in the same example.

  Doing this reveals that we actually have corrections which do not handle whitespace correctly. For now, the incorrect whitespace has been included in the specs, so they reflect the existing behavior.

  In addition, some examples are added covering edge cases, to capture the existing behavior ahead of changing it.

- **Refactors the `CallbackConditionalsBinding` cop**

  This rewrites the `CallbackConditionalsBinding` cop, making the following changes:

    - The cop no longer inherits from the deprecated `RuboCop::Cop::Cop` class.
    - The `CALLBACKS` constant is renamed to `RESTRICT_ON_SEND`, which allows RuboCop to use it to limit which methods `on_send` is called for.
    - Offense detection is simplified by leveraging a declarative node pattern describing offending conditional callbacks.
    - Correction is simplified by better leveraging the AST API.

- Changes the node to which `CallbackConditionalsBinding` offenses are added

  This changes which node `CallbackConditionalsBinding` offenses are attached to from the entire "macro" node to just the keyword argument pair node containing the `if:/unless:` keyword, and the offending block.

For changelog purposes, these changes could be summarized as

> - Improve handling of edge cases in `CallbackConditionalsBinding`.
> - Improve `CallbackConditionalsBinding` offense node accuracy.

### DIFF
--- a/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
+++ b/lib/rubocop/cop/sorbet/callback_conditionals_binding.rb
@@ -97,7 +97,7 @@ module RuboCop
 
             arg.each_child_node do |pair_node|
               argumentless_unbound_callable_callback_conditional?(pair_node) do |block|
-                add_offense(node, message: format(MSG, type: type)) do |corrector|
+                add_offense(pair_node, message: format(MSG, type: type)) do |corrector|
                   block_opening_indentation = block.source_range.source_line[/\A */]
                   block_body_indentation    = block_opening_indentation + SPACE * configured_indentation_width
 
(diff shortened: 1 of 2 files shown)


paracycle (approved):
This looks great, thank you!

vinistock (approved):

PR #209, opened 2024-03-24
Setup Dependabot

sambostock:
This configures Dependabot to bump gems and GitHub Actions versions.

See https://github.com/Shopify/rubocop-sorbet/pull/208#issuecomment-2016872194 for motivation (in addition to all the normal reasons we would want up-to-date dependencies).


andyw8:
Oh, wrong extension on the filename? `.config` instead of `.yml`.

PR #210, opened 2024-03-24
Bump minimum RuboCop version

sambostock:
_This PR originally just added the minimum RuboCop version to the CI matrix, but it was [discovered](https://github.com/Shopify/rubocop-sorbet/pull/210#issuecomment-2016910622) that that version doesn't work, so in now also bumps it._

This bumps the minimum RuboCop version to 1.45.1, to reflect actual compatibility.

It also adds the minimum RuboCop version to the CI matrix. This ensures we're compatible with the minimum RuboCop version we claim in `rubocop-sorbet.gemspec`.

This approach is [copied from `rubocop-shopify`](https://github.com/Shopify/ruby-style-guide/blob/main/gemfiles/minimum_rubocop.gemfile).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,10 @@ jobs:
       fail-fast: false
       matrix:
         ruby: ["2.7", "3.0", "3.1", "3.2"]
-    name: Test Ruby ${{ matrix.ruby }}
+        gemfile: ["Gemfile", "gemfiles/minimum_rubocop.gemfile"]
+    name: Test Ruby ${{ matrix.ruby }} | ${{ matrix.gemfile }}
+    env:
+      BUNDLE_GEMFILE: ${{ matrix.gemfile }}
     steps:
       - uses: actions/checkout@v4
       - name: Set up Ruby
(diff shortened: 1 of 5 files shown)


sambostock:
Actually testing against RuboCop 0.90.0 reveals that we don't support it (despite our gemspec saying we do).

#207 doesn't work as RuboCop 1.6 was the first version to define `RuboCop::ConfigObsoletion.files`, which we attempt to write to. I looked into backporting this, but 0.90.0 is so old it would be a huge change.

However, using RuboCop 1.6.0 still has over 100 test failures. I ran the test suite against every RuboCop version, and the point where it crosses over into compatibility is around RuboCop `1.45.1`.

RuboCop Version|Tests
-|-
`< 1.43.0`|❌ Fail
`= 1.43.0`|❌ Fail
`= 1.44.0`|✅ Pass
`= 1.44.1`|✅ Pass
`= 1.45.0`|❌ Fail
`= 1.45.1`|✅ Pass
`= 1.46.0`|✅ Pass
`> 1.46.0`|✅ Pass

Based on that, I propose that we change the minimum RuboCop version to reflect actual compatibility (rather than try to backport compatibility).

```diff
-spec.add_runtime_dependency("rubocop", ">= 0.90.0")
+spec.add_runtime_dependency("rubocop", ">= 1.45.1")
```

andyw8 (approved):

sambostock:
@andyw8 I'll address both your Ruby version related comments in a separate PR 👍

PR #222, opened 2024-04-10
Remove byebug and use debug gem instead

egiurleo:


st0012 (approved):

spec/spec_helper.rb
```diff
@@ -2,7 +2,6 @@
 
 require "rubocop-sorbet"
 require "rubocop/rspec/support"
-require "byebug"
```
vinistock, line 5: Should we add `require "debug/prelude"`, so that we can use breakpoints without requiring the debugger manually?

egiurleo: Oh I didn't know that was a thing! Yeah let's do it.

Issue #219, opened 2024-04-03
Discourage misuses of `T::Enum`


egiurleo:
While `T::Enum` provides a low-effort, type-safe way to construct an exhaustive set of values, it is also less performant than implementing the same functionality using data structures from Ruby (e.g. list of constants).

rubocop-sorbet should discourage using `T::Enum` in ways that exacerbates performance issues or doesn't use `T::Enum` for its intended purpose.

Potential cops include:
- [x] Convert any `T::Enum` with one value into a constant
- [x] Disallow including the `Comparable` module in `T::Enum` (comparing `T::Enum` values has particularly poor performance)
- [ ] ~~Disallow case statements with more than a certain number of `T::Enum` values ([motivation](https://github.com/sorbet/sorbet/issues/7532))~~

---

EDIT (2024-04-22): I decided not to create a cop forbidding case statements above a certain size.

After running some benchmarks, it seems that, at runtime, case statements have exponentially decreasing performance whether they're being used with `T::Enum` values or Ruby constants, and when statically type checking, time spent increases linearly. There isn't an obvious point at which performance drops off in either of these cases, making any cut off we assign totally arbitrary. We can revisit this issue if it comes up in the future.

<img width="972" alt="Screenshot 2024-04-22 at 4 25 43 PM" src="https://github.com/Shopify/rubocop-sorbet/assets/9601737/724a915f-c7ff-457e-9455-0badf297224d">

<img width="841" alt="Screenshot 2024-04-22 at 4 28 00 PM" src="https://github.com/Shopify/rubocop-sorbet/assets/9601737/af26ee48-8559-4e40-b717-be3f1e1866ca">

egiurleo:
@amomchilov That's really interesting! I will take a look at this as it ties in well with our Sorbet performance project. I wonder if we could apply the same logic to the `<=>` method? I did some benchmarks where `<`/`>` comparisons were 8x slower with `T::Enum` values than constants, which was pretty wild to me.

amomchilov:
@egiurleo `Comparable` in general isn't _great_ for performance, because it does a bunch of branching, error handling, etc. E.g.

* if you return a number greater than `1` or less than `-1`
* if you return `nil`, it'll raise an error
* etc.

[...]

PR #224, opened 2024-04-15
Introduce `MultipleTEnumValues` cop

egiurleo:
Related to #219.

This cop ensures that all `T::Enum`s have at least two values to limit unnecessary uses of `T::Enum`.

### DIFF
--- a/lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
+++ b/lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Disallow creating a `T::Enum` with less than two values.
+      #
+      # @example
+      #
+      #  # bad
+      #  class ErrorMessages < T::Enum
+      #    enums do
+      #      ServerError = new("There was a server error.")
+      #    end
+      #  end
+      #
+      #  # good
+      #  class ErrorMessages < T::Enum
+      #    enums do
+      #      ServerError = new("There was a server error.")
+      #      NotFound = new("The resource was not found.")
+      #    end
+      #  end
+      class MultipleTEnumValues < Base
+        def initialize(*)
+          @scopes = []
+          super
+        end
+
+        MSG = "`T::Enum` should have at least two values."
+
+        # @!method t_enum?(node)
+        def_node_matcher :t_enum?, <<~PATTERN
+          (class (const...) (const (const nil? :T) :Enum) ...)
+        PATTERN
+
+        # @!method enums_block?(node)
+        def_node_matcher :enums_block?, <<~PATTERN
+          (block (send nil? :enums) ...)
+        PATTERN
+
+        def on_class(node)
+          @scopes.push(node)
+
+          add_offense(node) if t_enum?(node) && node.body.nil?
+        end
+
+        def after_class(node)
+          @scopes.pop
+        end
+
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
+          return if @scopes.empty? || !t_enum?(@scopes.last)
+          return unless enums_block?(node)
+
+          scope = @scopes.last
+
+          if node.body.nil?
+            add_offense(scope)
+            return
+          end
+
+          begin_node = node.children.find(&:begin_type?)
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
```diff
@@ -0,0 +1,72 @@
...
+      #    enums do
+      #      ServerError = new("There was a server error.")
+      #    end
+      #  end
+      #
+      #  # good
+      #  class ErrorMessages
+      #    SERVER_ERROR = "There was a server error."
+      #  end
+
```
Morriar, line 21:
```suggestion
```

lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
```diff
@@ -0,0 +1,72 @@
...
+      #  class ErrorMessages < T::Enum
+      #    enums do
+      #      ServerError = new("There was a server error.")
+      #    end
+      #  end
+      #
+      #  # good
+      #  class ErrorMessages
+      #    SERVER_ERROR = "There was a server error."
+      #  end
```
Morriar, line 20: The `good` one is confusing in this case, as the cop do not really advise to use a constant instead. I think a better `good` here may be an enum with 2 values?

lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
```diff
@@ -0,0 +1,72 @@
...
+
+        def after_class(node)
+          @scopes.pop if t_enum?(node)
+        end
+
+        def on_block(node) # rubocop:disable InternalAffairs/NumblockHandler
+          return if @scopes.empty?
+          return unless enums_block?(node)
+
+          scope = @scopes.last
```
Morriar, line 52: If this only so we attach the offense to the class? Why not attaching it to the block instead?

lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
```diff
@@ -0,0 +1,72 @@
...
+            return
+          end
+
+          begin_node = node.children.find(&:begin_type?)
+
+          num_casgn_nodes = if begin_node
+            begin_node.children.count(&:casgn_type?)
+          else
+            node.children.count(&:casgn_type?)
+          end
```
Morriar, line 65: Not sure if this would be easier to read?

```suggestion
          num_casgn_nodes = (begin_node || node).children.count(&:casgn_type?)
```

Or something like this?

```suggestion
          node = begin_node if begin_node
          node.children.count(&:casgn_type?)
```

In any case, let's remove the `children.count(&:casgn_type?)` duplication.

spec/rubocop/cop/sorbet/multiple_t_enum_values_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      ^^^^^^^^^^^^^^^^^^^^^^ `T::Enum` should have at least two values.
+        class NestedEnum < T::Enum; end
+
+        enums do
+          A = B
+        end
+      end
+    RUBY
+  end
+end
```
Morriar, line 61: Would this break since we never reset the scope when entering a new class?

```rb
class MyEnum < T::Enum
  class Foo
    enums do; end
  end
end
```

Morriar: Hmm we may also want to check for this?

```rb
class MyEnum < T::Enum
end
```

In which case we would need to keep the offense on the class rather than the `enums` block.

egiurleo: For your first example, I think that would just be a method error because `enums` wouldn't exist on the `Foo` class?

Good catch on the second example.

egiurleo: Added a test for the second example.

lib/rubocop/cop/sorbet/multiple_t_enum_values.rb
egiurleo: I made the modifications you requested but kept the error on the class for now. Happy to talk more about it.

PR #228, opened 2024-04-29
Validate RBI config in CI

andyw8:
To prevent issues such as https://github.com/Shopify/rubocop-sorbet/pull/227


egiurleo:
I think I'm a bit confused about the approach here -- why don't we need to verify the other configs in this repo?

egiurleo (approved):

PR #235, opened 2024-05-30
Introduce `ForbidNestedTMust` cop

jacobobq:
This cop enforces replacing nested calls to `T.must` with a single call and safe-navigation operators.


jacobobq:
Hello! 👋🏼 If you are interested in this PR let me know any necessary steps to merge it.

I've written this for our codebase, mainly to help improve readability after using sorbet's automatic fixes.

jacobobq:
I have signed the CLA!

amomchilov (commented):
Hey Jacobo, thanks for taking time to implement this!

This is something I've corrected manually on many PRs, it would be really great to have RuboCop take care of it.

There's a "gotcha" with Ruby's safe-navigation operator: it's not actually a safe navigation operator (like from other languages). 😅

It's a "conditional send", Jez explains the difference in this blog post: [Ruby's Conditional Send is not Safe Navigation](https://blog.jez.io/conditional-send/)

I ran this against our monolith, and got many errors like this:

```
Method flatten does not exist on NilClass component of T.nilable(T::Array[Something]]) https://vault.shopify.io/page/Sorbet-error-reference~dhbd3ef.md#7003
     192 | T.must(input.surface_area_extension_points&.values.flatten),
                                                              ^^^^^^^
```

For this auto-correction to work correctly, the `.` needs to be replaced with `&.` not just on the first method call, but on every subsequent chained method, too.

amomchilov:
Here's an exact, unedited case you can test against:

```ruby
      assert_equal(
        I18n.t("sales.order_payments.errors.order_not_found"),
        T.must(T.must(result.error_value).errors.first).type
      )
```

It auto-corrected to:

```ruby
      assert_equal(
        I18n.t("sales.order_payments.errors.order_not_found"),
        T.must(result.error_value&.errors.first).type
      )
```

Which has this type-checking error:

```
file.rb:74: Used &. operator on ActiveModel::Errors, which can never be nil https://vault.shopify.io/page/Sorbet-error-reference~dhbd3ef.md#7034
    74 |      assert_equal(I18n.t("sales.flash.orders.saving_failed"), T.must(result.error_value&.errors.first).type)
```

lib/rubocop/cop/sorbet/forbid_nested_t_must.rb
```diff
@@ -0,0 +1,82 @@
...
+      #
+      #    # good
+      #    T.must(A.d(T.must(A.b&.c)))
+      class ForbidNestedTMust < Base
+        extend AutoCorrector
+        include IgnoredNode
+
+        MSG =
+          "Please avoid nesting `T.must` calls, instead use a " \
+            "single `T.must` around a safe-navigation operator chain"
```
amomchilov, line 38: What do you think of Jez's "conditional send" naming? It's fitting here, given that he's one of the primary contributors of Sorbet :D

```suggestion
            "single `T.must` around a conditional-send (`&.`) chain"
```

jacobobq: Sure! Let's use that instead! :)

amomchilov:
Hey Jacobo,

We discussed this internally, and found several reasons why we won’t be able to adopt this cop:
1. It can’t auto-correct on its own, and would lead to a poor experience for anyone using Rubocop auto-formatting, but not Sorbet auto-formatting.
2. There are [ongoing efforts](https://github.com/sorbet/sorbet/pull/7439) to improve the `T.must` error message (to tell you not just the line that raised, but which exact part of the line, so you can tell two `T.must` calls apart), which the `&.`-ified form wouldn’t benefit form.
3. While the nested `T.musts` are unsightly, they make a stronger guarantee about the expected values in the expression (that not just the final result can’t be `nil`, but the intermediate values in the chain can’t be nil, either). Switching to `&.` removes the guarantees about the intermediate values.
    1. Also, the auto-corrected form can’t be undone, because it’s impossible to automatically if a `&.` was hand-written, or a replacement for what used to be a nested `T.must`, so it’s a permanent loss of this information.

I’d encourage you to publish it under your own gem, perhaps the broader community might be interested!

Issue #238, opened 2024-07-10
New deprecation warning after upgrading rubocop to 1.65


mikdiet:
I'm getting this warning after rubocop updated to the latest.

```
/usr/local/bundle/gems/rubocop-sorbet-0.8.3/lib/rubocop/cop/sorbet/sigils/valid_sigil.rb:21: warning: `Cop.registry` is deprecated. Use `Registry.global` instead.
```

andyw8:
Thanks. I see this was introduced by https://github.com/rubocop/rubocop/pull/13032

MikeMcQuaid:
@andyw8 Any chance of a new release with the fix for this in it? The warning is hard to bypass and pretty annoying. Thanks for all your work on this useful cop ❤️

andyw8:
@MikeMcQuaid released: https://rubygems.org/gems/rubocop-sorbet/versions/0.8.4

PR #240, opened 2024-07-10
Fix registry deprecation notice

andyw8:
Closes https://github.com/Shopify/rubocop-sorbet/issues/238

### DIFF
--- a/lib/rubocop/cop/sorbet/sigils/has_sigil.rb
+++ b/lib/rubocop/cop/sorbet/sigils/has_sigil.rb
@@ -15,7 +15,7 @@ module RuboCop
       #
       # If a `MinimumStrictness` level is specified, it will be used in offense messages and autocorrect.
       class HasSigil < ValidSigil
-        @registry = Cop.registry # So we can properly subclass this cop
+        @registry = Registry.global # So we can properly subclass this cop
 
         def require_sigil_on_all_files?
           true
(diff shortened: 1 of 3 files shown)


Morriar (commented):
I think the right approach would be to migrate to `RuboCop::Cop::Base`. See https://docs.rubocop.org/rubocop/v1_upgrade_notes.html.

@sambostock started the transition with https://github.com/Shopify/rubocop-sorbet/pull/143.

andyw8:
Started work on moving to `RuboCop::Cop::Base`.

koic:
@andyw8 Yeah, I've opened #241 to solve it. Since `Cop.registry` warning is being addressed in this PR, it is not handled in that PR.

andyw8:
Thanks. I've reset this branch to contain only the `Cop.registry` change.

lib/rubocop/cop/sorbet/sigils/has_sigil.rb
```diff
@@ -15,7 +15,7 @@ module Sorbet
       #
       # If a `MinimumStrictness` level is specified, it will be used in offense messages and autocorrect.
       class HasSigil < ValidSigil
-        @registry = Cop.registry # So we can properly subclass this cop
+        @registry = Registry.global # So we can properly subclass this cop
```
paracycle, line 18: Is this even meaningful anymore? The comment seems to indicate that subclasses would not inherit the `@registry` instance variable and would thus break `Cop.registry` for subclass. Since we are moving to `Registry.global`, it feels like this should no longer be necessary.

lib/rubocop/cop/sorbet/sigils/valid_sigil.rb
```diff
@@ -18,7 +18,7 @@ module Sorbet
       # If an `ExactStrictness` level is specified, it will be used in offense messages and autocorrect.
       # Otherwise, if a `MinimumStrictness` level is specified, it will be used in offense messages and autocorrect.
       class ValidSigil < RuboCop::Cop::Cop # rubocop:todo InternalAffairs/InheritDeprecatedCopClass
-        @registry = Cop.registry # So we can properly subclass this cop
+        @registry = Registry.global # So we can properly subclass this cop
```
paracycle, line 21: Same as above.

PR #249, opened 2024-07-25
Enable `Layout/SpaceInsideReferenceBrackets` for RBIs

Morriar:
To avoid stuff like `T::Array[ Foo]`.

### DIFF
--- a/config/rbi.yml
+++ b/config/rbi.yml
@@ -175,6 +175,9 @@ Layout/SpaceInsideBlockBraces:
 Layout/SpaceInsideParens:
   Enabled: true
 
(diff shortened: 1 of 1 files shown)


amomchilov (approved):
Just to clarify, this profile only applies to RBI files in the repos of users who include `rubocop-sorbet`, right?

Should this perhaps live in https://github.com/Shopify/ruby-style-guide instead?

LGTM in any case

andyw8:
@amomchilov yeah you have to explicity enable the RBI config:

https://github.com/Shopify/rubocop-sorbet?tab=readme-ov-file#rubocop-rules-for-rbi-files

PR #251, opened 2024-07-29
Use `add_global_offense` for `ForbidRBIOutsideOfAllowedPaths`

Earlopain:
The previous `add_offense` errors when the file is empty: Range 0..1 makes no sense in that context. The error occurs during output:

```
test.rbi:1:1: C: Sorbet/ForbidRBIOutsideOfAllowedPaths: RBI file path should match one of: rbi/**, sorbet/rbi/**
index 0 outside of array bounds: 0...0
```

`add_global_offense` was added for Rubocop v1 which is great since 1.0.0 is now the minimum required version.

#### Why even bother?

I created a new rbi file for testing and `ruby-lsp` runs over it with rubocop immediatly which means it will error.

### DIFF
--- a/lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
+++ b/lib/rubocop/cop/sorbet/rbi/forbid_rbi_outside_of_allowed_paths.rb
@@ -27,16 +27,10 @@ module RuboCop
           paths = allowed_paths
 
           if paths.nil?
-            add_offense(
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
It makes sense to me to not have a location for these offenses.

andyw8:
This makes me realize that Ruby LSP doesn't have a way to show 'global' offenses:

https://github.com/Shopify/ruby-lsp/issues/2388

PR #252, opened 2024-07-30
Set `DocumentationBaseURL` and `DocumentationExtension`

Earlopain:
This allows `ruby-lsp` to link to cop documentation for this extension (notice the blue link which wasn't present before).

image

It takes you to https://github.com/Shopify/rubocop-sorbet/blob/main/manual/cops_sorbet.md#sorbetstrictsigil

Documentation here is in markdown and rubocop didn't support that until recently, so only set this when it supports both options. Otherwise it would link to non-existant html docs.

Requires a rubocop release with https://github.com/rubocop/rubocop/pull/13080, draft for now.

### DIFF
--- a/lib/rubocop/sorbet/inject.rb
+++ b/lib/rubocop/sorbet/inject.rb
@@ -11,6 +11,14 @@ module RuboCop
         def defaults!
           path = CONFIG_DEFAULT.to_s
           hash = ConfigLoader.send(:load_yaml_configuration, path)
+          if Gem::Version.new(RuboCop::Version::STRING) >= Gem::Version.new("1.66")
+            # We use markdown for cop documentation. Unconditionally setting
+            # the base url will produce incorrect urls on older RuboCop versions,
+            # so we do that for fully supported version only here.
+            hash["Sorbet"] ||= {}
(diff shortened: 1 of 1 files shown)


lib/rubocop/sorbet/inject.rb
```diff
@@ -11,6 +11,14 @@ class << self
         def defaults!
           path = CONFIG_DEFAULT.to_s
           hash = ConfigLoader.send(:load_yaml_configuration, path)
+          if RuboCop::Version::STRING >= "1.66"
```
amomchilov, line 14: Is this the customary way to check the Rubocop version? I'm worried it might not work in some edge cases (because it'll do string based ordering, not numerical ordering)

I asked about this in the Rubocop Discord: https://discord.com/channels/797752368217063424/797752368217063427/1280883199794483280

amomchilov (commented):
This is a great idea! I just have one q about the way the version check is done, but I love this change in principle

amomchilov (approved):

Earlopain:
@amomchilov  Let's continue this here. Automerge still seemed to be stuck, I rebased on latest main.

PR #256, opened 2024-08-20
Exclude unfreezable Sorbet invocations from Style/MutableConstant

dduugg:
With `strict` mode enabled, `Style/MutableConstant` requires constants declared with `T.type_alias` and `type_member` to be frozen. However, doing so results in Sorbet [errors](https://github.com/sorbet/sorbet/issues/3532).

Thus, I would like to exclude these unfreezable patterns from `Style/MutableConstant`.

### DIFF
--- a/lib/rubocop/cop/sorbet/mutable_constant_sorbet_aware_behaviour.rb
+++ b/lib/rubocop/cop/sorbet/mutable_constant_sorbet_aware_behaviour.rb
@@ -12,6 +12,12 @@ module RuboCop
             base.def_node_matcher(:t_let, <<~PATTERN)
               (send (const nil? :T) :let $_constant _type)
             PATTERN
+            base.def_node_matcher(:t_type_alias?, <<~PATTERN)
+              (block (send (const {nil? cbase} :T) :type_alias ...) ...)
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/mutable_constant_sorbet_aware_behaviour.rb
```diff
@@ -12,13 +12,23 @@ def prepended(base)
             base.def_node_matcher(:t_let, <<~PATTERN)
               (send (const nil? :T) :let $_constant _type)
             PATTERN
+            # @!method t_type_alias?(node)
+            base.def_node_matcher(:t_type_alias?, <<~PATTERN)
+              (block (send (const {nil? cbase} :T) :type_alias ...) ...)
+            PATTERN
+            # @!method type_member?(node)
```
dduugg, line 19: Just FYI, I dropped this YARD tag here as required by rubocop, but we've had good results with this tapioca compiler: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/sorbet/tapioca/compilers/rubocop.rb

PR #258, opened 2024-08-23
Add `ForbidTEnum` cop

Morriar:
So we can deprecate usages of `T::Enum`.


lib/rubocop/cop/sorbet/forbid_t_enum.rb
```diff
@@ -0,0 +1,44 @@
...
+      #   # bad
+      #   class MyEnum < T::Struct
+      #     enums do
+      #       A = new
+      #       B = new
+      #     end
+      #   end
+      #
+      #   # good
+      #   class MyEnum
```
Morriar, line 21: I'm not convinced it's the best way to represent this? Maybe @rafaelfranca has a better idea?

vinistock: This `good` section feels odd to me. Do we want to advise people to write their own enums like this one?

Or maybe I'm missing context. What are people intended to use in replacement for the enum?

vinistock (approved):
Code looks good to me, just wondering about the recommendation for fixing the violations

lib/rubocop/cop/sorbet/forbid_t_enum.rb
rafaelfranca: > But then adding a custom class to act as an enum would have the same drawbacks as using `T::Enum`, right? Even a bit worse because people would have to add their own serializers as needed.

Yes, exactly, that is why I don't want people to use custom classes. Just use the raw values.

PR #282, opened 2025-02-19
Tweak Rubocop rules for inline RBS in RBI files

Morriar:
So we can use RBS comments within RBI files (especially useful to write shims):

```rb
#: -> String
def to_s; end
```


Morriar:
> Can you change in the ruby-style-guide instead? [Shopify/ruby-style-guide#688](https://github.com/Shopify/ruby-style-guide/pull/688)

@rafaelfranca I opened a PR against your branch: https://github.com/Shopify/ruby-style-guide/pull/693 👍

PR #289, opened 2025-02-26
Pluginfy RuboCop Performance

ngan:


lib/rubocop/sorbet/plugin.rb
```diff
@@ -0,0 +1,31 @@
...
+require "lint_roller"
+
+module RuboCop
+  module Sorbet
+    # A plugin that integrates RuboCop Sorbet with RuboCop's plugin system.
+    class Plugin < LintRoller::Plugin
+      def about
+        LintRoller::About.new(
+          name: 'rubocop-sorbet',
+          version: Version::STRING,
```
Morriar, line 12: Shouldn't this be pointing to `Sorbet::VERSION`?

ngan: 👍🏼

paracycle (commented):
The changes look good to me, but I guess this will mean that 0.9 and above of this gem will only work with RuboCop 1.72.1 and above.

I wish there was a gradual pathway to doing this so that we didn't have to drop support for older versions of RuboCop. 1.72 is very new and not everyone may be ready to upgrade to that.

Can we not keep the janky inject code, and gate it on the RuboCop version to change behaviour?

ngan:
> Can we not keep the janky inject code, and gate it on the RuboCop version to change behaviour?

@paracycle done

Morriar:
There are a few lint errors, looks like they can all be autocorrect with `bundle exec rubocop -a`.

Morriar (approved):
Thanks!

PR #291, opened 2025-02-28
[WIP] Update Sorbet/EnforceSignatures to accept RBS signatures

bitwise-aiden:
### Description

This is a first pass at updating `Sorbet/EnforceSignatures` to accept RBS signatures. 

Some open questions that I'd like to answer: 
- Should this be it's own `Sorbet/EnforceRbsSignatures`?
  - If not should we add the ability to configure `sig` | RBS | both?
- How do we want to determine if the comment is a signature?
  - At the moment it is just looking for `#: ...` to keep inline with not checking for validity of present `sig`s. 
  - Should we support multiline now?
- Is there any enforcement of proximity to method, i.e. can there be comments before / after sig: 
``` ruby
# comment before 
#: -> void 
# comment after
def some_method; end
```


lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
```diff
@@ -76,6 +78,17 @@ def check_node(node)
           @last_sig_for_scope[scope] = nil
         end
 
+        def has_rbs_comment?(node)
+          node = node.parent while RuboCop::AST::SendNode === node.parent
```
bitwise-aiden, line 82: For the case of visibility / methods such as: 
```ruby 
private def some_method; end
foo bar baz def some_method; end
```
Is there a better way of doing this?

andyw8:
> Should this be it's own Sorbet/EnforceRbsSignatures?
> If not should we add the ability to configure sig | RBS | both?

How about `Sorbet/EnforceSignatureFormat`, with options for `sig` and `rbs`. No need for `both` since that would be allowed by default if you disable the rule.

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
KaanOzkan, line 1: Thanks. I think it'll be good to have a test for a new line between the comment and the method def:
```
#: -> String

def foo; end
```

KaanOzkan (commented):
>Should this be it's own Sorbet/EnforceRbsSignatures?
If not should we add the ability to configure sig | RBS | both?

I prefer them together since you can use them together in a codebase. Configuration options can be used to specify if a repo should only accept RBS or Sig.

> How do we want to determine if the comment is a signature?
At the moment it is just looking for #: ... to keep inline with not checking for validity of present sigs.

That's fine

> Should we support multiline now?
I think no for now, syntax could change

> Is there any enforcement of proximity to method, i.e. can there be comments before / after sig:

As long as `#:` is part of the node it should be type checked. So I think you can have comments between method and the RBS comment but not a new line.

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
Morriar: ^ Yes, let's add a test for this.

Currently Sorbet does not associate the RBS sig to the method is there is a blank line in between.

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
```diff
@@ -28,6 +28,8 @@ class EnforceSignatures < ::RuboCop::Cop::Base
         extend AutoCorrector
         include SignatureHelp
 
+        RBS_COMMENT_REGEX = /^#:.*$/
```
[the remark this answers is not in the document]
Morriar: Seems about right: https://github.com/rubocop/rubocop/blob/323cb71dce89b9700decd508cbd5c2594876d4c2/lib/rubocop/cop/layout/leading_comment_space.rb#L178

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
Morriar: Would it make it easier to put the comment into `@last_sig_for_scope[scope]` as we visit the parent nodes?

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
```diff
@@ -215,6 +312,22 @@ class Foo
...
+        protected def foo; end
+
+        #: -> void
+        foo bar baz def foo; end
+      RUBY
+    end
+
     shared_examples_for("autocorrect with config") do
       it("autocorrects methods by adding signature stubs") do
         expect(
```
Morriar, line 333: Let's add your example with comments:

```rb
# some comment
#: -> void
# some other comment
def foo; end
```

is a valid sig.

But this isn't (for now) because of the blank line:

```rb
#: -> void

def foo; end
```

Morriar (commented):
This should be configurable as an option on the cop like `AllowRBS: true` and disabled by default.

bitwise-aiden:
> How about Sorbet/EnforceSignatureFormat, with options for sig and rbs. No need for both since that would be allowed by default if you disable the rule.

@andyw8 Right, so this cop will check either are present, then adding that cop to say it also needs to be x format?

> I prefer them together since you can use them together in a codebase. Configuration options can be used to specify if a repo should only accept RBS or Sig.

@KaanOzkan 👍🏻 I like @andyw8's suggestion of a separate rule to enforce the style of the signature. 

> This should be configurable as an option on the cop like AllowRBS: true and disabled by default.

@Morriar Would that be as well as the suggestion @andyw8 made, or instead of?

spec/rubocop/cop/sorbet/signatures/enforce_signatures_spec.rb
bitwise-aiden: Before adding this. Should this be in this cop or a different cop?

With the info I have I could do: 
``` ruby 
def has_rbs_comment?(node)
  node = node.parent while RuboCop::AST::SendNode === node.parent
  return false unless (comments = preceeding_comments(node))

  last_comment = comments.last
  return false comment.loc.line + 1 < node.loc.line

  comments.any? { |comment| RBS_COMMENT_REGEX.match?(comment.text) }
end
```

But that would currently attach the `Each method is required to have a signature.` error. Would it be more fitting to add another cop that specifically validates that there is no space for a more useful message like: `There must be no blank line between an RBS signature and a method`.

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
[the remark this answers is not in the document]
bitwise-aiden: `/#:|#\[.+\]/`

Does this mean that you could have `#[sometext]` as a valid RBS signature? If so I should probably update it.

lib/rubocop/cop/sorbet/signatures/enforce_signatures.rb
Morriar: Okay, let's keep the recursion on receivers 👍

Morriar:
> This should be configurable as an option on the cop like AllowRBS: true and disabled by default.

This needs to be done in this cop.

The RBS comments are only interpreted by Sorbet if passing `--enable-experimental-rbs-signatures` which is not the default behaviour.

PR #298, opened 2025-03-17
Add `Sorbet/AnonymousClassBlock` cop

magni-:
This is to avoid running into https://github.com/sorbet/sorbet/issues/3609

### DIFF
--- a/lib/rubocop/cop/sorbet/anonymous_class_block.rb
+++ b/lib/rubocop/cop/sorbet/anonymous_class_block.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require "rubocop"
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Disallow calling `Class.new` with a block.
+      #
+      # Sorbet incorrectly assumes any methods or constants defined in these blocks
+      # are privately defined on `Object`, which can lead to false positives when type-
+      # checking (Sorbet may think that a method is defined when it isn't).
+      # See https://github.com/sorbet/sorbet/issues/3609#issuecomment-727137772:
+      # > Sorbet is somewhat fundamentally incompatible with anonymous classes,
+      # > and [Sorbet recommends] avoiding them.
+      #
+      # @example
+      #   # bad
+      #   Class.new do
+      #     def this_is_bad
+      #     end
+      #   end
+      #
+      #   # good
+      #   class NamedClass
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/sorbet/anonymous_class_block.rb
```diff
@@ -0,0 +1,59 @@
...
+          return unless class_new_with_block?(node)
+
+          message =
+            if (alternative = cop_config["Alternative"])
+              "#{MSG} Use `#{alternative}` instead."
+            else
+              MSG
+            end
+
+          add_offense(node, message: message)
```
magni-, line 45: This line was initially written using Ruby 3.1 syntax (`add_offense(node, message:)`) which caused the `lint` job to fail because Rubocop sets `TargetRubyVersion` from the `.gemspec`, which currently has `spec.required_ruby_version = ">= 3.0"`. This change can be reverted if #299 is merged.

magni-:
> Doesn't this cast a too wide of a net due to support for assignments with `Class.new`?https://github.com/sorbet/sorbet/blob/master/rewriter/ClassNew.h
> 
> I think we should only disallow the ones without assignment.
> 
> Also I'm wary of enabling this by default without an auto correct, I'm looking at some usages in our repo and in most cases it'll add errors for potential future type checking errors for code that works.

Ah good catch, I'd misunderstood the range of the bug, methods defined in `Class.new` don't pollute `Object` as long as `Class.new` is assigned to a constant 👍🏼 

[sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0AClass.new%20do%0A%20%20def%20_anon%3B%20end%20%0Aend%0A%0Aa%20%3D%20Class.new%20do%0A%20%20def%20_var%3B%20end%0Aend%0A%0AA%20%3D%20Class.new%20do%0A%20%20def%20_const%3B%20end%0Aend%0A%0Aclass%20B%0A%20%20def%20named%0A%20%20%20%20_anon%0A%20%20%20%20_var%0A%20%20%20%20_const%0A%20%20end%0Aend%0A)
```ruby
Class.new do
  def _anon; end 
end

a = Class.new do
  def _var; end
end

A = Class.new do
  def _const; end
end

class B
  def named
    _anon  # No error
    _var   # No error
    _const # Error: Method _const does not exist on B
  end
end
```

PR #301, opened 2025-03-25
Fix plugin/require logic

paracycle:
There were a few things that weren't right about supporting plugin mode optionally:
1. We should never have a declared dependency on `lint_roller` since it is supposed to be brought in by RuboCop.
2. We can just return early without loading the plugin code at all (since without a dependency on LintRoller, it wouldn't even load) and then check if `::RuboCop::Sorbet::Plugin` is defined , to decide to activate the old logic.
3. We should also not be opinionated about how the library is loaded in the `config/rbi.yml` file. For most clients, `rubocop-sorbet` would have been loaded (either by `require` or `plugins`) in the top-level `.rubocop.yml` file, so no need to do anything in our YAML file. If they are not doing that, then all they would have to do is to add the relevant `require` or `plugins` line to the `sorbet/rbi/.rubocop.yml` file manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       # to a directory that doesn't contain any .rb or .rbi files.
       - name: Validate default config
         run: |
-          bin/rubocop --config config/rbi.yml config
+          bin/rubocop --plugin rubocop-sorbet --config config/rbi.yml config
           bin/rubocop --config config/default.yml config -r rubocop-sorbet
       - name: Lint Ruby files
         run: bin/rubocop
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):
Thanks

lib/rubocop/sorbet/plugin.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+rubocop_min_version = Gem::Version.create("1.72.1")
+rubocop_version = Gem::Version.create(RuboCop::Version.version)
+
+return if rubocop_version < rubocop_min_version
+
 require "rubocop"
 require "lint_roller"
```
rafaelfranca, line 9: If this is required here, we need to declare a dependency on it. Even if it is just a `gem "lint_roller"` in this file so `lint_roller` is a optional dependency. If Rubocop remove the lint_roller dependecy,  this gem will just fail to load.

PR #310, opened 2025-05-06
Translate all test from RSpec to Minitest

rafaelfranca:
We should not use RSpec anymore since it isn't the default test runner for Shopify projects.

Most conversion was done with help of the AI agent.

We were using RSpec before because there was no Minitest support in RuboCop, but now that we have it, we should use it.

Test suite now finishes in 1.2s instead of 2.5s with RSpec, but I don't think that is a fair comparison because there are less tests.


Morriar (approved):
I haven't compared all the files but the structure looks okay

PR #312, opened 2025-05-09
Add cop to suggest using `grep` over `select` when using it only for type narrowing

### LINKED ISSUE #247: Add cop to suggest `grep` over `select` when using only for type narrowing
Sorbet now supports proper type inference when narrowing types using `grep` [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Astrings_or_integers%20%3D%20T.let%28%5B%5D%2C%20T%3A%3AArray%5BT.any%28String%2C%20Integer%29%5D%29%0A%0AT.reveal_type%28strings_or_integers.grep%28String%29%29).

This is a great improvement to ergonomics because it avoids having to cast the type after having already narrowed the elements of a collection. This is currently only supported for `grep`, since Sorbet can be sure of the behaviour, whereas `select` is more complex and allows for other conditions.

Let's add a cop that suggest using `grep` when `select` is being used exclusively for type narrowing:

```ruby
strings_or_integers = T.let([], T::Array[T.any(String, Integer)])

# Bad. Sorbet cannot infer the right types when using `select`
strings_or_integers.select { |e| e.is_a?(String) }

# Good. The right types are inferred with `grep`
strings_or_integers.grep(String)
```

rafaelfranca:
Since sorbet can narrow down the type of the receive when using `grep` we should be suggesting that instead of `select`.

Closes #247.


test/rubocop/cop/sorbet/select_by_is_a_test.rb
```diff
@@ -0,0 +1,189 @@
...
+
+module RuboCop
+  module Cop
+    module Sorbet
+      class SelectByIsATest < ::Minitest::Test
+        MSG = "Use `grep` instead of `select` when using it only for type narrowing."
+
+        def setup
+          @old_parser_engine = ENV["PARSER_ENGINE"]
+          ENV["PARSER_ENGINE"] = "parser_prism"
```
vinistock, line 13: Why do we need to switch the parser engine?

config/default.yml
```diff
@@ -349,3 +349,12 @@ Sorbet/ForbidComparableTEnum:
   Description: 'Disallows including the `Comparable` module in a `T::Enum`.'
   Enabled: true
   VersionAdded: 0.8.2
+
+Sorbet/SelectByIsA:
+  Description: >-
+    Suggests using `grep` over `select` when using it only for type narrowing.
+    This is because Sorbet can properly infer types when using `grep` but not with `select`.
+  Enabled: pending
```
vinistock, line 357: Should we just go with `true` straight away?

vinistock (approved):

config/default.yml
rafaelfranca: We can, but that isn't how historically rubocop plugins works. They consider enabling cops breaking changes so they require major version bumps. That is why they introduce cops as pending in minor versions. But we are pre 1.0, so I have no problem on making breaking changes

PR #330, opened 2025-06-04
Bump rubocop requirement to include `any_def_type?`

KaanOzkan:
Resolves `undefined method 'any_def_type?' for an instance of RuboCop::AST::DefNode` errors when using an older rubocop version.

Introduced in [v1.75.2](https://github.com/rubocop/rubocop/commit/ae1f33c406cce91e979c9e46ceb2a03dafce97ed)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     rubocop-sorbet (0.10.1)
       lint_roller
-      rubocop (>= 1)
+      rubocop (>= 1.75.2)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


KaanOzkan:
> Can we stop using any_def_type? instead of forcing downstream users to use a 2 month old release of RuboCop?

Did some more digging. `any_def` was added to the internal affairs cop in the same version [here](https://github.com/rubocop/rubocop/commit/ae1f33c406cce91e979c9e46ceb2a03dafce97ed). So we got a [violation](https://github.com/Shopify/rubocop-sorbet/actions/runs/14578445983/job/40889649798#step:5:15) when we bumped our rubocop version for using `def/defs`.

Looks like this was intended by `rubocop` but we could lower the development Rubocop version we use in `rubocop-sorbet` and keep `def/defs`. Or change usage depending on the `rubocop` version.

rafaelfranca:
Ah, we are using `any_def`

rafaelfranca:
I'd not bother. 2 months old rubocop is old enough. if people want to keep using old rubocop, they should also use old rubocop-sorbet

rafaelfranca (approved):

PR #335, opened 2025-06-09
Fix a few issues with `BlockMethodDefinition` cop

### LINKED ISSUE #334: BlockMethodDefinition rewrites methods defined inside of a ActiveSupport::Concern `class_methods` blocks
`ActiveSupport::Concern` has special support in Sorbet's system, but the new Rubocop check introduced by #319 doesn't recognize it and rewrites method definitions to use `define_method`, breaking Sorbet's detection of them.

For example,

```ruby
module Example
  extend ActiveSupport::Concern

  class_methods do
    extend T::Sig

    sig { params(a: Numeric, b: Numeric).void }
    def add(a, b)
      @result = T.let(a + b, Numeric)
    end
  end
end
```

This works just fine when included into a class elsewhere, with Sorbet correctly detecting the presence of a new `add` class method, but the new Rubocop rule complains about it, with the autocorrect changing it to use `define_method` instead, after which Sorbet no longer finds it and then it, in turn, complains about calls to missing methods.

### LINKED ISSUE #337: BlockMethodDefinition autocorrection incorrect for heredocs
Version: 0.10.2

Input:

```ruby
      def example
        <<~XML
          abc
        XML
      end
```

Autocorrection:

```ruby
      define_method(:example) do ||
        <<~XML
      end
```

### LINKED ISSUE #338: BlockMethodDefinition autocorrection removes comments
Input:

```ruby
      def example
        # abc
      end
```

Correction:

```ruby
      define_method(:example) do
      end
```

rafaelfranca:
Fixes #334.
Fixes #337.
Fixes #338.

### DIFF
--- a/lib/rubocop/cop/sorbet/block_method_definition.rb
+++ b/lib/rubocop/cop/sorbet/block_method_definition.rb
@@ -107,8 +107,7 @@ module RuboCop
           indent = offset(node)
 
           method_name = node.method_name
-          args = node.arguments.map(&:source).join(", ")
-          args = " |#{args}|" unless args.empty?
+          args = transform_args_to_block_args(node)
 
           # Build the method signature replacement
           if node.def_type?
@@ -128,6 +127,17 @@ module RuboCop
           signature_range = node.source_range.with(end_pos: end_pos)
 
           corrector.replace(signature_range, signature_replacement + indentation)
+        end
+
+        def transform_args_to_block_args(node)
+          args = node.arguments
+
+          if args.empty?
+            ""
+          else
+            args_string = args.map(&:source).join(", ")
+            " |#{args_string}|"
+          end
         end
 
         def handle_method_without_body(node, indent)


lib/rubocop/cop/sorbet/block_method_definition.rb
```diff
@@ -61,6 +95,15 @@ def on_block(node)
 
         private
 
+        def in_activesupport_concern_class_methods_block?(node)
+          return false unless activesupport_concern_class_methods_block?(node)
+
+          immediate_module = node.each_ancestor(:module).first
+          return false unless immediate_module
```
KaanOzkan, line 102: What does this test for given the 2 node matchers?

rafaelfranca: hmmm, maybe I don't need it. Let me write a test and remove.

rafaelfranca:
Now I included more fixes for the other 3 issues with this cops

KaanOzkan (approved):

PR #336, opened 2025-06-09
Handle methods with no arguments in `BlockMethodDefinition` autocorrector

magni-:
I missed this edge case when adding the cop in #319 🙇🏼


Morriar (approved):

Morriar:
Thanks!

PR #345, opened 2025-06-25
Add CI check for stale version placeholders

### LINKED ISSUE #343: Automate releases or prevent releasing without running the release rake task
Current release steps consists of updating the VERSION file and running `rake prepare_release`, https://github.com/Shopify/rubocop-sorbet/blob/main/CONTRIBUTING.md#releasing-a-new-version

Rake task is a unique flow in our step and can be missed. We can either:

### 1. Automate the release
  - This can be done through a GitHub actions workflow, instead of manual bumping we trigger the workflow that runs all necessary commands 
  - Workflow can take a version number as input or the strings `major`, `minor`, `patch`
### 2. Check if rake task has ran before a release
  - We could check in CI before a deploy that rake task has successfully updated the necessary files
    - No `<<next>>` usages in `config/default.yml`
    - Updated `lib/rubocop/sorbet/version.rb`
  - This check will need to be triggered only on new releases and not normal merges to `main`

st0012:
- Add rake task to verify <<next>> placeholders are replaced when VERSION changes
- Add GitHub Actions workflow to run check on main branch pushes
- Prevents releasing with unreplaced version placeholders in `config/default.yml`

Closes #343

### DIFF
--- a/.github/workflows/check-version-placeholders.yml
+++ b/.github/workflows/check-version-placeholders.yml
@@ -1,0 +1,16 @@
+name: Check Version Placeholders
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  check-placeholders:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Check for stale version placeholders
+        run: bundle exec rake check_version_placeholders
--- a/Rakefile
+++ b/Rakefile
@@ -47,10 +47,6 @@ module Releaser
     version = File.read("VERSION").strip
     puts "Preparing release for version #{version}"
 
-    update_file("lib/rubocop/sorbet/version.rb") do |version_file|
-      version_file.sub(/VERSION = ".*"/, "VERSION = \"#{version}\"")
-    end
-
     update_file("config/default.yml") do |default|
       default.gsub(/['"]?<<\s*next\s*>>['"]?/i, "'#{version}'")
     end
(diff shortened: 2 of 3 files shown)


Rakefile
```diff
@@ -73,3 +73,25 @@ module Releaser
     File.write(path, yield(content))
   end
 end
+
+desc "Check for stale <<next>> placeholders when VERSION is updated"
+task :check_version_placeholders do
+  # Check if VERSION file was modified in the last commit
+  version_changed = %x(git diff HEAD~1 HEAD --name-only).split("\n").include?("VERSION")
```
vinistock, line 80: Should we check `lib/rubocop/sorbet/version.rb` too in case someone bumps it directly?

.github/workflows/check-version-placeholders.yml
```diff
@@ -0,0 +1,20 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  check-placeholders:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
vinistock, line 13: If we're always checking the latest commit, why do we need this?

Rakefile
vinistock: I actually like having the `VERSION` constant exist, so I'll suggest that we read the content from the file. What do you think?

st0012: That works too. Updated.

lib/rubocop/sorbet/version.rb
```diff
@@ -2,6 +2,6 @@
 
 module RuboCop
   module Sorbet
-    VERSION = "0.10.5"
+    VERSION = File.read(File.expand_path("../../../VERSION", __dir__)).strip
```
rafaelfranca, line 5: We should not reading reading a file when loading this file just to get a string. Why do you need this?

st0012: 2 things need the version: the gem's `VERSION` constant and the `prepare_release` rake task. So we have both this file and the `VERSION` file as the version source.
And since we're adding a check on version change, I think it's better to have just one source of truth.
Comparing with parsing `version.rb`'s `VERSION` constant, IMO it's easier to directly read from the `VERSION` file, so I decided to use it as the single version source.

PR #349, opened 2025-08-26
Fix docs url for lsp

Earlopain:
Since this extension was made into a plugin, the change from https://github.com/Shopify/rubocop-sorbet/pull/252 does not take anymore, since the inject-based codepath is not taken.

To fix this, simply move the necessary config into the yaml file since the required rubocop version now allows it. 
I also removed some dead code now that the required rubocop version is so high in the second commit


test/project_test.rb
```diff
@@ -50,7 +50,7 @@ def config
     @config ||= RuboCop::ConfigLoader.load_file("config/default.yml")
   end
 
-  def cops_on(config)
-    config.keys.reject { |key| key == "inherit_mode" }
+  def sorbet_cops(config)
+    config.keys.select { |key| key.start_with?("Sorbet/") }
```
Earlopain, line 54: Required to filter out top-level `Sorbet` (which is not a cop)

PR #358, opened 2025-12-05
Check that `on_failure` is used with `checked(:{tests,always})`

issyl0:
- Our app had a PR that made it into our pre-production environment with Sorbet signatures that used `on_failure` without `checked(:tests)` or `checked(:always)`.
- Hence, we got an error that we had to rollback for, and I thought it would be a good candidate for a cop to catch this statically.

```
Exception: ArgumentError
Details: Error interpreting sig:
 To use .on_failure you must additionally call .checked(:tests) or .checked(:always), otherwise, the .on_failure has no effect.
```

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked.rb
+++ b/lib/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Checks that `on_failure` is not used without `checked(:tests)` or `checked(:always)`.
+      #
+      # @example
+      #
+      #   # bad
+      #   sig { params(x: Integer).returns(Integer).on_failure(:raise) }
+      #   def plus_one(x)
+      #     x + 1
+      #   end
+      #
+      #   # good
+      #   sig { params(x: Integer).returns(Integer).checked(:always).on_failure(:raise) }
+      #   def plus_one(x)
+      #     x + 1
+      #   end
+      #
+      class RuntimeOnFailureDependsOnChecked < ::RuboCop::Cop::Base
+        include SignatureHelp
+
+        MSG = "To use .on_failure you must additionally call .checked(:tests) or .checked(:always), otherwise, the .on_failure has no effect."
+
+        # @!method on_failure_call?(node)
+        def_node_matcher :on_failure_call?, <<~PATTERN
+          (send _ :on_failure ...)
+        PATTERN
+
+        # @!method checked_tests_or_always?(node)
+        def_node_matcher :checked_tests_or_always?, <<~PATTERN
+          (send _ :checked (sym {:tests | :always}))
+        PATTERN
+
+        def on_signature(node)
+          return unless node.descendants.any? { |n| on_failure_call?(n) }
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Checks that `on_failure` is not used without `checked(:tests)` or `checked(:always)`.
```
Morriar, line 6: Can you add a good/bad example as we do in other cops?

See https://github.com/Shopify/rubocop-sorbet/blob/main/lib/rubocop/cop/sorbet/forbid_t_untyped.rb for example.

lib/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        def_node_matcher :checked_tests_or_always?, <<~PATTERN
+          (send _ :checked (sym {:tests | :always}))
+        PATTERN
+
+        def on_signature(node)
+          return unless node.descendants.any? { |n| on_failure_call?(n) }
+          return if node.descendants.any? { |n| checked_tests_or_always?(n) }
+
+          add_offense(node, message: MSG)
```
Morriar: Using `MSG` is already the default behaviour:

```suggestion
          add_offense(node)
```

test/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked_test.rb
```diff
@@ -0,0 +1,46 @@
...
+              sig { void.checked(:always).on_failure(:raise) }
+            RUBY
+          end
+
+          def test_ok_with_only_checked
+            assert_no_offenses(<<~RUBY)
+              sig { params(x: Integer).returns(Integer).checked(:always) }
+            RUBY
+          end
+
```
Morriar:
```suggestion
```

test/rubocop/cop/sorbet/signatures/runtime_on_failure_depends_on_checked_test.rb
```diff
@@ -0,0 +1,46 @@
...
+            @cop = RuntimeOnFailureDependsOnChecked.new
+          end
+
+          def test_offends_when_on_failure_is_without_checked_always_or_tests
+            assert_offense(<<~RUBY)
+              sig { params(x: Integer).returns(Integer).on_failure(:raise) }
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{MSG}
+
+              sig { void.on_failure(:log) }
+              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{MSG}
```
Morriar, line 34: Can you add a few tests with `on_failure` without args `checked()`, `checked(true)`, `checked("tests"), etc. please?

issyl0:
Thanks for the review, @Morriar! I pushed the changes.

Morriar:
You have a few lint errors but it's otherwise okay 👍

Morriar (approved):

PR #362, opened 2026-01-20
Add RedundantTLet cop

### LINKED ISSUE #346: Suggested cop: remove T.let when its signature matches the param signature in initialize
Prevent T.let from duplicating the information in params for initialize methods since Sorbet will infer the type without the T.let.

Bad:

```ruby
sig { params(a: Integer) }
def initialize(a)
  @a = T.let(a, Integer)
end
```

Good:

```ruby
sig { params(a: Integer) }
def initialize(a)
  @a = a
end
```

Allowed (any time the let definition is different than the param definition):

```ruby
sig { params(a: Integer) }
def initialize(a)
  @a = T.let(a, T.any(Integer, String))
end
```

dejmedus:
- closes https://github.com/Shopify/rubocop-sorbet/issues/346

Sorbet [can infer the type](https://github.com/Shopify/sorbet/blob/master/rewriter/Initializer.h) of `initialize` method instance variables when they are assigned to parameters already typed by signatures. In [cases](https://sorbet.run/#%20%23%20typed%3A%20true%0A%20%0A%20class%20Bar%0A%20%20%20%20extend%20T%3A%3ASig%0A%0A%20%20%20%20sig%20%7B%20params%28b%3A%20Integer%29.void%20%7D%0A%20%20%20%20def%20initialize%28b%29%0A%20%20%20%20%20%20%40b%20%3D%20b%0A%20%20%20%20end%0A%0A%20%20%20%20sig%20%7B%20void%20%7D%0A%20%20%20%20def%20use_b%0A%20%20%20%20%20%20T.reveal_type%28%40b%29%0A%20%20%20%20end%0A%0A%0A%20%20%20%20sig%20%7B%20params%28a%3A%20Integer%29.void%20%7D%0A%20%20%20%20def%20set_a%28a%29%0A%20%20%20%20%20%20%40a%20%3D%20a%0A%20%20%20%20end%0A%0A%20%20%20%20sig%20%7B%20void%20%7D%0A%20%20%20%20def%20use_a%0A%20%20%20%20%20%20T.reveal_type%28%40a%29%20%0A%20%20%20%20end%0A%20%20end) where a `T.let` assertion is:

- top level inside an `initialize` method
- assigned to an instance variable
- typed the same as in `sig` params

we can DRY up code by removing it

``` diff
sig { params(foo: String) }
def initialize(foo:)
- @foo = T.let(foo, String)
+ @foo = foo
end
```

The cop will display `Sorbet/RedundantTLet: Unnecessary T.let. The instance variable type is inferred from the signature` and optionally autocorrect

<details><summary>Demo</summary>

https://github.com/user-attachments/assets/f62ead5b-a07b-4fe0-b7f5-36302b45b714

</details>


test/rubocop/cop/sorbet/redundant_t_let_test.rb
```diff
@@ -0,0 +1,208 @@
...
+              @a = T.let(a, T.nilable(Integer))
+                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{MSG}
+            end
+
+            sig { params(a: T::Array[Integer]).void }
+            def initialize(a)
+              @a = T.let(a, T::Array[Integer])
+                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{MSG}
+            end
+          RUBY
```
Morriar, line 89: Let's add a few example with more complicated type syntax:

```
sig { params(a: Foo[Bar], b: T.any(A, B), c: T.proc.void).void }
def initialize(proc)
  @a = T.let(a, Foo[Foo]) # error
  @b = T.let(b, T.any(B, A)) # it's equivalent, should it be ok or error?
  @c = T.let(c, T.proc.returns(Integer)) # ok
end
```

Morriar (commented):
This is looking good, let's add a few tests to encode current behaviour with more complex type expressions then we can merge it 👍

test/rubocop/cop/sorbet/redundant_t_let_test.rb
dejmedus: Added a [few more examples](https://github.com/Shopify/rubocop-sorbet/pull/362/changes#diff-ccff86d56fcd552ffcf20139f190c7ef7a03520a621ba7bbbfc568cce95e973fR55-R89):) 

> @ b = T.let(b, T.any(B, A)) # it's equivalent, should it be ok or error?

Right now the cop is checking for syntactical equivalence, not semantical, so [`T.any(A, B)` and `T.any(B, A)`](https://github.com/Shopify/rubocop-sorbet/pull/362/changes#diff-ccff86d56fcd552ffcf20139f190c7ef7a03520a621ba7bbbfc568cce95e973fR60-R62) (or `T.proc.params(x: A, y: B)` and `T.proc.params(y: B, x: A)`) are treated as different and left alone

I think it would be possible to manually check for parity in these cases but the solutions I looked at felt a bit fragile. I would lean towards not warning/autocorrecting. What do you think?

Morriar (commented):
<deleted>

KaanOzkan (approved):
LGTM

Morriar (approved):

PR #367, opened 2026-02-09
Add new `Sorbet/RedundantTLetForLiteral` cop

dduugg:
Similar to `Sorbet/RedundantTLet` (#362), which removes redundant `T.let` in `initialize` methods where instance variable types can be inferred from the signature, this cop targets a different case: constant assignments with simple literals.

When a constant is assigned a simple literal (Float, Integer, Regexp, String, Symbol) wrapped in `T.let` with the matching class as the type, the `T.let` is redundant because Sorbet can infer the type automatically. For example:

```ruby
# bad — Sorbet already knows this is Integer
MAX_RETRIES = T.let(3, Integer)

# good
MAX_RETRIES = 3
```

The cop only fires on constant assignments (`casgn`) with simple literals. It correctly ignores collections, nilable types, instance variables, local variables, and cases where the type doesn't match the literal.

Includes autocorrection support.


lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
```diff
@@ -0,0 +1,73 @@
...
+          float: :Float,
+          int: :Integer,
+          regexp: :Regexp,
+          str: :String,
+          sym: :Symbol,
+        }.freeze
+
+        def on_casgn(node)
+          t_let_with_literal_and_class?(node) do |literal_node, class_name|
+            next unless LITERAL_TYPE_TO_CLASS[literal_node.type] == class_name
```
amomchilov, line 63: I think you can use [`RuboCop::AST::Node#literal?`](https://github.com/rubocop/rubocop-ast/blob/2bcb8189d54d7bd54fb71ac34b5a554289fda07d/lib/rubocop/ast/node.rb#L431-L433), which covers a few more cases

Though I guess you still need to map it to the class name

test/rubocop/cop/sorbet/redundant_t_let_for_literal_test.rb
```diff
@@ -0,0 +1,296 @@
...
+
+        def test_no_offense_for_rational_literal
+          assert_no_offenses(<<~RUBY)
+            VALUE = T.let(0.3r, Rational)
+          RUBY
+        end
+
+        def test_no_offense_for_frozen_string
+          assert_no_offenses(<<~RUBY)
+            VALUE = T.let("hello".freeze, String)
```
Morriar, line 200: Let's also add one for string interpolation:

```rb
FOO = T.let("#{42}", String)
```

test/rubocop/cop/sorbet/redundant_t_let_for_literal_test.rb
```diff
@@ -0,0 +1,296 @@
...
+
+        def test_no_offense_for_boolean_false
+          assert_no_offenses(<<~RUBY)
+            FLAG = T.let(false, T::Boolean)
+          RUBY
+        end
+
+        def test_no_offense_for_nil
+          assert_no_offenses(<<~RUBY)
+            VALUE = T.let(nil, T.nilable(String))
```
Morriar, line 218: Let's also add one where we widen the type to a non-nilable:

```rb
FOO = T.let(42, Numeric)
```

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
```diff
@@ -0,0 +1,73 @@
...
+        PATTERN
+
+        # Maps AST node types to the class name Sorbet would infer.
+        LITERAL_TYPE_TO_CLASS = {
+          float: :Float,
+          int: :Integer,
+          regexp: :Regexp,
+          str: :String,
+          sym: :Symbol,
+        }.freeze
```
Morriar, line 59: Arrays of literals are also [accepted](https://sorbet.run/#%23%20typed%3A%20strict%0A%0AFOO%20%3D%20%5B1%2C%202.0%2C%20%223%22%5D%0A):

```rb
FOO = [1, 2.0, "3"]
```

But not 0-length tuples:

```rb
FOO = [] # error: Constants must have type annotations with `T.let` when specifying `# typed: strict`
```

test/rubocop/cop/sorbet/redundant_t_let_for_literal_test.rb
dduugg: Good catch, looks like there's a missing literal type to support (`dstring`), will update

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
dduugg: Are you ok with punting on this? My rough draft roughly triples the content of the cop, and it would only apply to unfrozen array constants, which are banned in the codebases I work in.

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
dduugg: The issue is that `literal?` is too [broad](https://github.com/rubocop/rubocop-ast/blob/2bcb8189d54d7bd54fb71ac34b5a554289fda07d/lib/rubocop/ast/node.rb#L30-L42). It includes arrays, hashes, and ranges we don't handle—while `basic_literal?` is too narrow since we handle `dstr` and `regexp`. I think the best approach is to use `literal?` as an initial filter in the node matcher, then verify the specific type matches what we support in the `LITERAL_TYPE_TO_CLASS` map. It's a bit redundant but helps clarify intent.

dduugg:
@amomchilov @Morriar Thanks for the initial feedback, please take a look at the latest commit.

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
```diff
@@ -0,0 +1,74 @@
...
+        MSG = "Redundant `T.let` for %{type} literal. Sorbet can infer this type automatically."
+
+        # @!method t_let_with_literal_and_class?(node)
+        def_node_matcher :t_let_with_literal_and_class?, <<~PATTERN
+          (casgn _ _ (send (const nil? :T) :let $literal? (const nil? $_)))
+        PATTERN
+
+        # Maps AST literal node types to the class name Sorbet would infer.
+        LITERAL_TYPE_TO_CLASS = {
+          dstr: :String,
```
amomchilov, line 53: `dsym` as well

```ruby
def rubocop_ast(source)
  RuboCop::AST::ProcessedSource.new(source, RUBY_VERSION.to_f).ast
end

rubocop_ast ':"interpolated #{123} symbol"'

s(:dsym,
  s(:str, "interpolated "),
  s(:begin,
    s(:int, 123)),
  s(:str, " symbol"))
```

Morriar (commented):
<deleted>

dduugg:
@Morriar I'm not sure what to make of your comment. This repo doesn't use `sorbet-runtime`, I don't understand the suggestion to add `sig`s. I don't see `system` used anywhere, let alone in the changes introduced in this PR.

I think I'm going to hand this off to y'all at this point, if this is something you're interested in. It has been used to indentify and autofix > 1000 violations in a repo without error or false positive.

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
Morriar: Fine by me 👍

Morriar:
> @Morriar I'm not sure what to make of your comment.

Nevermind, it's a rogue AI impersonating me! 😅

amomchilov:
Rebased and fixed the merge-conflict

PR #372, opened 2026-04-13
Fix Sorbet/RedundantTLet: match T.let type against multiline sig annotations

dduugg:
## Summary

Fixes one bug in the `Sorbet/RedundantTLet` cop introduced in #362,
and adds tests documenting two edge cases where T.let must be kept.

### Bug: Multiline sig type annotations are not matched

When a sig uses a multiline type annotation, `sig_type.source` contains
newlines, indentation, and trailing commas that don't appear in the
single-line T.let argument, causing a spurious no-match:

```ruby
sig do
  params(
    a: T.any(
      Integer,
      String,
    ),
  ).void
end
def initialize(a)
  @a = T.let(a, T.any(Integer, String))  # was not flagged
end
```

Fix: normalize both sides of the type comparison by collapsing
whitespace and stripping trailing commas before closing delimiters.

### Documented edge case: method modifier-wrapped def

When `initialize` is wrapped by a method modifier (`private def initialize`),
Sorbet's initializer rewriter does **not** process ivar assignments, so
`T.let` remains required for type correctness at `# typed: strict`.
The cop correctly skips these methods by checking whether the def's
parent is a send node.

Note: `initialize` is already private, so `private def initialize` is
unusual, but any other method modifier would trigger the same issue.

### Documented edge case: rescue in initialize body

`initialize` bodies with a method-level `rescue` clause are correctly
left alone. Sorbet's initializer rewriter does not process ivar
assignments inside a rescue-wrapped body, so removing `T.let` there
would cause a type error in `# typed: strict` files. The existing
`ivar_assignments` guard already handles this, but the behavior was
not previously tested.

### DIFF
--- a/lib/rubocop/cop/sorbet/redundant_t_let.rb
+++ b/lib/rubocop/cop/sorbet/redundant_t_let.rb
@@ -61,7 +61,7 @@ module RuboCop
           # When the def is wrapped by a method modifier (`private def initialize`),
           # Sorbet's initializer rewriter does not process the ivar assignments,
           # so T.let annotations remain required. Skip by returning nil.
-          return nil if method_node.parent&.send_type?
+          return if method_node.parent&.send_type?
 
           method_node.left_sibling.then { |s| signature?(s) ? s : nil }
         end
--- a/lib/rubocop/cop/sorbet/redundant_t_let.rb
+++ b/lib/rubocop/cop/sorbet/redundant_t_let.rb
@@ -42,8 +42,8 @@ module RuboCop
           method_args = node.arguments&.to_h { |arg| [arg.name, arg.type] }
           return unless method_args&.any?
 
-          sig_node = node.left_sibling
-          return unless sig_node && signature?(sig_node)
+          sig_node = find_sig_node(node)
+          return unless sig_node
 
           sig_params = sig_params(sig_node)&.to_h { |pair| [pair.key.value, pair.value] }
           return unless sig_params&.any?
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/redundant_t_let.rb
```diff
@@ -57,6 +57,26 @@ def on_def(node)
 
         private
 
+        def find_sig_node(method_node)
+          # When the def is wrapped by a method modifier (`private def initialize`),
+          # Sorbet's initializer rewriter does not process the ivar assignments,
+          # so T.let annotations remain required. Skip by returning nil.
+          return nil if method_node.parent&.send_type?
```
dejmedus: Thank you for catching this!

CI has a tiny [linting nit](https://github.com/Shopify/rubocop-sorbet/actions/runs/24351554399/job/71309981141?pr=372):

```suggestion
          return if method_node.parent&.send_type?
```

dejmedus (approved):
Thanks very much! This is great :)

lib/rubocop/cop/sorbet/redundant_t_let.rb
```diff
@@ -57,6 +57,26 @@ def on_def(node)
...
+        end
+
+        def normalize_whitespace(source)
+          source
+            .gsub(/\s+/, " ")              # collapse all whitespace to single spaces
+            .gsub(/,\s*([)\]\}])/, "\\1")  # remove trailing commas before closing delimiters
+            .gsub(/\(\s*/, "(")            # remove space after (
+            .gsub(/\s*\)/, ")")            # remove space before )
+            .gsub(/\[\s*/, "[")            # remove space after [
+            .gsub(/\s*\]/, "]")            # remove space before ]
```
dejmedus, line 76: Question I missed, what do you think about also removing space after commas, allowing something like `T.any(Integer,String)` to match `T.any(Integer, String)`? 

```suggestion
            .gsub(/\s*\]/, "]")            # remove space before ]
            .gsub(/,\s*/, ", ")            # remove space after comma
```

dduugg:
Thanks for the merge, @dejmedus! My PR queue got the best of me before I could address your comment. 🙇

Issue #373, opened 2026-04-15
Feature: Add AllowUntypedHashValues configuration option to Sorbet/ForbidTUntyped


dduugg:
## Summary

The `Sorbet/ForbidTUntyped` cop currently has no way to allow `T.untyped` in specific contexts. A common legitimate use case is using `T.untyped` as the **value type** of `T::Hash` (e.g., `T::Hash[Symbol, T.untyped]`) when the hash values are truly heterogeneous and cannot be more precisely typed. Notably, `T.untyped` as the **key type** (e.g., `T::Hash[T.untyped, String]`) should remain flagged.

## Proposed Configuration Option

Add an `AllowUntypedHashValues` boolean option (defaulting to `false` to preserve current behavior):

```yaml
Sorbet/ForbidTUntyped:
  Enabled: true
  AllowUntypedHashValues: true
```

With this configuration:

```ruby
# good (T.untyped in value position of T::Hash)
sig { returns(T::Hash[Symbol, T.untyped]) }
def metadata; end

# bad (T.untyped in key position of T::Hash — still flagged)
sig { returns(T::Hash[T.untyped, String]) }
def metadata; end

# bad (T.untyped used directly, not as a hash value type — still flagged)
sig { params(x: T.untyped).void }
def foo(x); end
```

## Motivation

It is common in large codebases to have hash-typed values where the value type is genuinely heterogeneous (e.g., JSON payloads, serialized attributes, plugin registries). For these cases, `T::Hash[Symbol, T.untyped]` is the idiomatic Sorbet type, and enforcing a non-untyped value type would require either:

- Using `T.untyped` for the entire hash (worse), or
- Disabling `ForbidTUntyped` entirely (defeats the purpose), or
- Using `# rubocop:disable` inline suppression comments everywhere (noisy).

`AllowUntypedHashValues` lets teams adopt `Sorbet/ForbidTUntyped` broadly while making a targeted exception only for the value position of `T::Hash`, where `T.untyped` is genuinely appropriate. The key type remains restricted, since an untyped key is rarely intentional.

## Implementation Notes

The cop would need to detect when a `T.untyped` node appears specifically as the **second** type argument (index 1) to `T::Hash[...]`. In the AST, `T::Hash[K, V]` is represented as a `send` on `T::Hash` with `[]` and two arguments — the cop should only skip the offense when `T.untyped` is in the second argument position.

The default value for `AllowUntypedHashValues` should be `false` (preserving current behavior).

PR #374, opened 2026-04-15
Add `AllowUntypedHashValues` option to `Sorbet/ForbidTUntyped`

dduugg:
## Summary

Implements the feature requested in #373.

Adds an `AllowUntypedHashValues` boolean configuration option (defaulting to `false`) to `Sorbet/ForbidTUntyped`. When enabled, `T.untyped` in the value position of `T::Hash` is allowed, while the key position and all other usages remain flagged.

```yaml
Sorbet/ForbidTUntyped:
  Enabled: true
  AllowUntypedHashValues: true
```

```ruby
# good (T.untyped in value position of T::Hash)
sig { returns(T::Hash[Symbol, T.untyped]) }
def metadata; end

# bad (T.untyped in key position — still flagged)
sig { returns(T::Hash[T.untyped, String]) }
def metadata; end

# bad (T.untyped used directly — still flagged)
sig { params(x: T.untyped).void }
def foo(x); end
```

## Test plan

- [x] Existing `ForbidTUntypedTest` tests continue to pass (default behavior unchanged)
- [x] `ForbidTUntypedAllowUntypedHashValuesTest` covers:
  - Value position allowed: `T::Hash[Symbol, T.untyped]` → no offense
  - Key position flagged: `T::Hash[T.untyped, String]` → offense
  - Both positions untyped: `T::Hash[T.untyped, T.untyped]` → offense on key only (not value)
  - Nested hash value position: `T::Hash[Symbol, T::Hash[String, T.untyped]]` → no offense
  - Nilable-wrapped hash: `T.nilable(T::Hash[Symbol, T.untyped])` → no offense
  - Union type containing hash: `T.any(T::Hash[Symbol, T.untyped], NilClass)` → no offense
  - Direct usage flagged: `T.untyped` → offense
  - `T::Array[T.untyped]` → offense
- [x] `ForbidTUntypedDefaultConfigTest` verifies `T::Hash[Symbol, T.untyped]` is flagged when `AllowUntypedHashValues` is `false` (default)

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -236,6 +236,7 @@ Sorbet/ForbidTUnsafe:
 Sorbet/ForbidTUntyped:
   Description: 'Forbid usage of T.untyped'
   Enabled: false
+  AllowUntypedHashValues: false
   VersionAdded: 0.6.9
   VersionChanged: 0.7.0
 
--- a/config/default.yml
+++ b/config/default.yml
@@ -238,7 +238,7 @@ Sorbet/ForbidTUntyped:
   Enabled: false
   AllowUntypedHashValues: false
   VersionAdded: 0.6.9
-  VersionChanged: 0.7.0
+  VersionChanged: '<<next>>'
 
 Sorbet/ForbidTypeAliasedShapes:
   Description: 'Forbids defining type aliases that contain shapes'
--- a/lib/rubocop/cop/sorbet/forbid_t_untyped.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_untyped.rb
@@ -59,7 +59,7 @@ module RuboCop
         def hash_value_position?(node)
           parent = node.parent
           parent&.send_type? &&
-            parent.method_name == :[] &&
+            parent.method?(:[]) &&
             t_hash_receiver?(parent.receiver) &&
             parent.arguments[1].equal?(node)
         end
--- a/lib/rubocop/cop/sorbet/forbid_t_untyped.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_untyped.rb
@@ -17,6 +17,22 @@ module RuboCop
       #   sig { params(my_argument: String).void }
       #   def foo(my_argument); end
       #
+      # @example AllowUntypedHashValues: false (default)
+      #
+      #   # bad
+      #   sig { returns(T::Hash[Symbol, T.untyped]) }
+      #   def metadata; end
+      #
+      # @example AllowUntypedHashValues: true
+      #
+      #   # good (T.untyped in value position of T::Hash)
+      #   sig { returns(T::Hash[Symbol, T.untyped]) }
+      #   def metadata; end
+      #
+      #   # bad (T.untyped in key position of T::Hash — still flagged)
+      #   sig { returns(T::Hash[T.untyped, String]) }
+      #   def metadata; end
+      #
       class ForbidTUntyped < RuboCop::Cop::Base
         MSG = "Do not use `T.untyped`."
         RESTRICT_ON_SEND = [:untyped].freeze
@@ -24,8 +40,28 @@ module RuboCop
         # @!method t_untyped?(node)
         def_node_matcher(:t_untyped?, "(send (const nil? :T) :untyped)")
 
+        # @!method t_hash_receiver?(node)
+        def_node_matcher(:t_hash_receiver?, "(const (const {nil? cbase} :T) :Hash)")
+
         def on_send(node)
-          add_offense(node) if t_untyped?(node)
+          return unless t_untyped?(node)
+          return if allow_untyped_hash_values? && hash_value_position?(node)
+
+          add_offense(node)
+        end
+
+        private
+
+        def allow_untyped_hash_values?
+          cop_config.fetch("AllowUntypedHashValues", false)
+        end
+
+        def hash_value_position?(node)
+          parent = node.parent
+          parent&.send_type? &&
+            parent.method_name == :[] &&
+            t_hash_receiver?(parent.receiver) &&
+            parent.arguments[1].equal?(node)
         end
       end
     end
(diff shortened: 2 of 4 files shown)


config/default.yml
```diff
@@ -236,6 +236,7 @@ Sorbet/ForbidTUnsafe:
 Sorbet/ForbidTUntyped:
   Description: 'Forbid usage of T.untyped'
   Enabled: false
+  AllowUntypedHashValues: false
   VersionAdded: 0.6.9
   VersionChanged: 0.7.0
```
KaanOzkan:
```suggestion
  VersionChanged: '<<next>>'
```

KaanOzkan (commented):
Thanks. I think this will encourage the use of `ForbidTUntyped`, PR looks good but I'll check in with the team to verify.

KaanOzkan:
@dduugg we don't really use this cop internally so I'm curious about your thinking for this feature instead of typing the value as `T.anything`. Is it the higher effort requirement?

amomchilov:
Hey @dduugg, I totally get the problem you're trying to address, but I'm not sold on this solution approach.

It's oddly specific to the config/metadata/etc. cases, which IMO is common enough to be worth a solution, but even still I don't think this one would be sufficient. Many of these use cases use a tree of not just Hashes, but also Arrays. So we'd need something like a `AllowUntypedArrayValues` option, too.

More generally, I think this hits in a weird spot. `Sorbet/ForbidTUntyped` is a pretty hardcore rule to enable, which selects for a particularly typing-devoted audience. I would posit that audience wouldn't be particularly fond of untyped Hash values, which leaves very little target audience for this feature. Put graphically:

Screenshot 2026-04-16 at 3 36 23 PM

I think `T.anything` would be a better way to model this behaviour. Hashes like `T::Hash[Symbol, T.anything]` can be passed around the system as-is, but at the point the values are used, you're forced to assert the type of the value you expect via a cast. I think this is more aligned with the users of `Sorbet/ForbidTUntyped`, and even more users beyond them:

Screenshot 2026-04-16 at 3 36 43 PM


Perhaps the biggest issue of `T.anything` (like `unknown` in TypeScript) is that it's just not very well known. Perhaps the `Sorbet/ForbidTUntyped` cop could help here, by offering an auto-correct from `T.untyped` to `T.anything`.

What do you think?

dduugg:
> @dduugg we don't really use this cop internally so I'm curious about your thinking for this feature instead of typing the value as `T.anything`. Is it the higher effort requirement?

@KaanOzkan I tried this in https://github.com/Homebrew/brew. Claude reported making 382 such replacements resulting in 838 type errors. I can paste excerpts, if you'd like more detail (or open a draft PR for your to inspect).

dduugg:
@amomchilov this is exactly the standard Homebrew is aiming for (which is to avoid T.untyped, except as a substitute for union value types in a typed hash, to avoid excessive type guards for every access)

amomchilov:
@dduugg sorry, what's the connection to Homebrew? Is it trying to onboard onto Sorbet?

What do you think about the suggested `T.anything` alternative?

dduugg:
> @dduugg sorry, what's the connection to Homebrew? Is it trying to onboard onto Sorbet?
> 
> What do you think about the suggested `T.anything` alternative?

@amomchilov https://github.com/Homebrew/brew is entirely `# typed: strict` 🔒, though we want to further ratchet `T.untyped` use.

See https://github.com/Shopify/rubocop-sorbet/pull/374#issuecomment-4262923504 for the impact of using `T.anything` for Hash value types as a replacement for `T.untyped` as a Hash value type.

amomchilov:
@dduugg Oh I missed that, I didn't see a `sorbet/` folder in the repo root, so I didn't think it was using it yet (just found `Library/Homebrew/sorbet/`). It's really cool to see such a big project using Sorbet, especially with such high strictness. Bravo!

> I tried [using `T.anything`] in https://github.com/Homebrew/brew. Claude reported making 382 such replacements resulting in 838 type errors. 

If you ask Claude to fix the type errors, how much does it end up negatively contorting your code to make it pass type-checking? I imagine the vast majority of required changes are added casts. Is it reasonable, or totally gross?

<hr>

For any of these cases, one could of course just slap on `# rubocop:disable Sorbet/ForbidTUntyped`, or add it to the YAML exclusion list. I guess the real question here is: is `T::Hash[_, T.anything]` a common enough pattern to warrant its own exclusion setting/pattern, over just excluding each declaration individually?

amomchilov:
@dduugg Another alternative that's less strict than `T.anything`: How well does using `Object` (or perhaps `BasicObject` work?

dduugg:
> @dduugg Another alternative that's less strict than `T.anything`: How well does using `Object` (or perhaps `BasicObject` work?

@amomchilov thanks for the suggestion. There's also shape types, T::Struct, and so on. I'm very familiar with sorbet, rubocop, and homebrew code, and `T::Hash[Symbol, T.untyped]` is the preferred steady state for the foreseable future, at least while we work on the punch list of `T.untyped` elsewhere. Your graphic makes a lot of sense for newer projects, but it's helpful for 15+ to have a transition path.

No worries if y'all don't want to support this, `brew` implements a number of custom cops, and this could be another one of them (although I have other codebases that i'm responsible for, some of which also have a place for this).

Thanks, all, for the consideration and discussion.

amomchilov:
@dduugg I think my only issue with this feature is its overly-specific focus on Hashes. It's fine on its own, but I wouldn't want to grow into a series of different `AllowUntypedXXXValues`.

How difficult would it be to design this to take a list of patterns, where you could do something like:

```yaml
Sorbet/ForbidTUntyped:
  Enabled: true
  AllowedPatterns:
    - 'T::Hash[_, T.untyped]`
```

> There's also shape types, T::Struct, and so on.

Good point

dduugg:
Thanks again for the thoughtful discussion and suggestions. Based on the feedback, I think it makes the most sense to close this out and implement it as one of Homebrew's custom cops.

PR #383, opened 2026-07-10
Allow configuring which style to use when autocorrecting signatures in `Sorbet/EnforceSignatures` cop

rafaelfranca:
When `Style: both` is configured, we were always autocorrecting to `sig` style. If your codebase is migration to RBS you want the autocorrection to be in RBS style. This change adds a new configuration option `AutocorrectStyle` which can be set to either `sig` or `rbs` to control which style is used when autocorrecting.

### DIFF
--- a/test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
+++ b/test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
@@ -428,26 +428,26 @@ module RuboCop
 
           def test_autocorrects_with_custom_values
             @cop = target_cop.new(cop_config({
-              "Style" => "both",
+              "Style" => "sig",
               "ParameterTypePlaceholder" => "PARAM",
               "ReturnTypePlaceholder" => "RET",
             }))
 
             assert_offense(<<~RUBY)
               def foo; end
-              ^^^^^^^^^^^^ #{MSG}
+              ^^^^^^^^^^^^ #{MSG_SIG}
               def bar(a, b = 2, c: Foo.new); end
-              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ #{MSG}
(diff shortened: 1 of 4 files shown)


test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
```diff
@@ -428,26 +428,26 @@ def foo; end
 
           def test_autocorrects_with_custom_values
             @cop = target_cop.new(cop_config({
-              "Style" => "both",
+              "Style" => "sig",
```
paracycle, line 431: I am not sure why these had to change? Seems irrelevant?

test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
```diff
@@ -473,18 +473,18 @@ def bar(a, b, c)
 
           def test_autocorrects_accessors_with_custom_values
             @cop = target_cop.new(cop_config({
-              "Style" => "both",
+              "Style" => "sig",
```
paracycle, line 476: Same comment as above

paracycle (approved):
Looks good to me, but not sure why some tests changed.

test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
rafaelfranca: Just to avoid having to change this test if we decide to change the default autocorrection to rbs in the future. Style is irrelevant for those test. I didn't remove it because `Stye` default can also change.

I initially changed it because I made `rbs` the default for the autocorrect and saw I had to change existing tests, so I change the tests in a way that changing the value of `Style` and `AutocorrectStyle` would not require any unrelated test to be changed.

PR #386, opened 2026-07-13
Fix Sorbet/EnforceSignatures to generate correct RBS signature comments for attr_*

rafaelfranca:
Before it was generating `#: () -> untyped` that is incorrect.

### DIFF
--- a/test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
+++ b/test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
@@ -232,11 +232,11 @@ module RuboCop
           def test_makes_no_offense_if_accessor_has_rbs_signature
             assert_no_offenses(<<~RUBY)
               class Foo
-                #: -> String
+                #: String
                 attr_reader :foo
-                #: (String) -> void
+                #: String
                 attr_writer :bar
-                #: (String) -> String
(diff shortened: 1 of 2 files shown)


test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
```diff
@@ -321,7 +321,7 @@ class Foo
           def test_does_not_check_rbs_signature_for_accessors
             assert_no_offenses(<<~RUBY)
               class Foo
-                #: -> void
+                #: void
```
paracycle: Curious: what does `void` mean for an attribute?

paracycle (approved):

test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb
rafaelfranca: This is just to test that the cop doesn't register an offence in an already existing signature. I asked the same question. This test should probably be changes to be more realistic.

vinistock: Yeah, I'd say we fix it. Void for an attribute is confusing

vinistock (approved):

PR #387, opened 2026-07-17
Fix TLetCorrection corrupting files with interior heredocs

dduugg:
### Problem

`TLetCorrection` (shared by `Sorbet/RedundantTLet` and `Sorbet/RedundantTLetForLiteral`) corrupts files when the `T.let` value contains an **interior** heredoc — one that is followed by more of the value's own source.

The mixin was written for heredocs at the *tail* of the value, where `value_node.source` stops at the marker and the body must be reattached. But when a heredoc sits in the interior of a larger value (e.g. a constructor call that continues after the heredoc), `value_node.source` already contains the heredoc body inline, so reattaching it duplicates the body.

Minimal repro:

```ruby
PATH = T.let(
  Foo.new(
    a: <<~DIR,
      /usr/local
    DIR
    b: 2,
  ),
  Foo,
)
```

Autocorrect currently produces a duplicated, syntactically broken body:

```ruby
PATH = Foo.new(
    a: <<~DIR,
      /usr/local
    DIR
    b: 2,
  )
      /usr/local   # <- duplicated
    DIR            # <- duplicated
```

### Fix

Only reattach heredocs whose body/terminator extend **past** the value node's own source range (the tail heredocs). Interior heredocs already live inside `value_node.source`, so the plain `corrector.replace(t_let_node, value_node.source)` path handles them correctly.

### Tests

Added `test_offense_on_constant_constructor_with_interior_heredoc` covering the case above. Full suite green (`bundle exec rake test`).

### DIFF
--- a/lib/rubocop/cop/sorbet/mixin/t_let_correction.rb
+++ b/lib/rubocop/cop/sorbet/mixin/t_let_correction.rb
@@ -13,10 +13,10 @@ module RuboCop
 
         # Replaces `T.let(value, Type)` with `value`.
         def replace_t_let(corrector, t_let_node, value_node)
-          heredocs = tail_heredocs(value_node)
-          return corrector.replace(t_let_node, value_node.source) if heredocs.empty?
+          heredocs_to_reattach = tail_heredocs(value_node)
+          return corrector.replace(t_let_node, value_node.source) if heredocs_to_reattach.empty?
 
-          replace_t_let_preserving_heredocs(corrector, t_let_node, value_node, heredocs)
+          replace_t_let_preserving_heredocs(corrector, t_let_node, value_node, heredocs_to_reattach)
         end
 
         # Heredocs at the tail of the value, whose body sits past `value_node`'s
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/mixin/t_let_correction.rb
```diff
@@ -13,12 +13,22 @@ module TLetCorrection
 
         # Replaces `T.let(value, Type)` with `value`.
         def replace_t_let(corrector, t_let_node, value_node)
-          heredocs = value_node.each_node(:any_str).select(&:heredoc?)
+          heredocs = tail_heredocs(value_node)
           return corrector.replace(t_let_node, value_node.source) if heredocs.empty?
```
KaanOzkan: Nit: `if heredocs.empty?` reads funny. Can we either change the variable name (`heredocs_to_reattach`?) or add a comment to indicate these heredocs require special handling.

dduugg: Renamed to `heredocs_to_reattach` in c319e15 — agreed, that reads much clearer. Thanks!

KaanOzkan (approved):

PR #389, opened 2026-07-28
Add `Sorbet/SetterReturnType` cop and extract RBSParser helper

rafaelfranca:
Add a new cop that flags setter methods (names ending in `=`, including `[]=` and Unicode identifiers, but excluding comparison operators like `==`, `<=`, `>=`, `!=`, `===`) whose signature does not declare a `void` return type. It supports both Sorbet `sig` blocks and RBS inline-comment signatures, with autocorrect that rewrites the return type to `void` (`SafeAutoCorrect: false` since the return contract changes).

Extract the RBS inline-comment parsing into a pure `RuboCop::Sorbet::RBSParser` module (used directly, not mixed in) whose public API is `rbs_signatures_before` plus the `Signature` value object exposing `return_type`, `return_type_range`, and `void?`. `EnforceSignatures::RBSSignatureChecker` now delegates to it.

### DIFF
--- a/lib/rubocop/cop/sorbet/signatures/setter_return_type.rb
+++ b/lib/rubocop/cop/sorbet/signatures/setter_return_type.rb
@@ -1,0 +1,155 @@
+# frozen_string_literal: true
+
+module RuboCop
+  module Cop
+    module Sorbet
+      # Checks that setter methods (methods whose name ends with `=`)
+      # declare a `void` return type, in either a Sorbet `sig` block or an
+      # RBS inline comment signature.
+      #
+      # Sorbet requires setter methods to return `void`. Declaring any other
+      # return type violates the setter contract and is rejected by Sorbet
+      # under `#typed: strict` and above.
+      #
+      # @example
+      #
+      #   # bad
+      #   sig { params(name: String).returns(String) }
+      #   def name=(name); end
+      #
+      #   # good
+      #   sig { params(name: String).void }
+      #   def name=(name); end
+      #
+      #   # bad
+      #   #: (String) -> String
+      #   def name=(name); end
+      #
+      #   # good
+      #   #: (String) -> void
+      #   def name=(name); end
+      #
+      #   # bad
+      #   #: (
+      #   #|   String name
+      #   #| ) -> String
+      #   def name=(name); end
+      #
+      #   # good
+      #   #: (
+      #   #|   String name
+      #   #| ) -> void
+      #   def name=(name); end
+      #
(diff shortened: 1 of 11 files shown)


lib/rubocop/cop/sorbet/signatures/setter_return_type.rb
```diff
@@ -0,0 +1,153 @@
...
+        extend AutoCorrector
+
+        MSG = "Setter methods must declare a `void` return type."
+
+        NON_SETTER_OPERATORS = ["==", "===", "!=", "<=", ">="].freeze
+
+        def on_def(node)
+          check(node)
+        end
+        alias_method :on_defs, :on_def
```
KaanOzkan, line 56: Should we enforce `attr_writer` sigs too?

lib/rubocop/cop/sorbet/signatures/setter_return_type.rb
```diff
@@ -0,0 +1,153 @@
...
+
+        def check_sig(sig)
+          decl = return_declaration(sig.body)
+          return unless decl # incomplete signature: nothing declared to check
+          return if decl.method?(:void)
+
+          return unless decl.first_argument # malformed returns
+
+          add_offense(sig_return_range(decl)) do |corrector|
+            corrector.replace(sig_return_range(decl), "void")
```
KaanOzkan: `sig_return_range(decl)` can be called once and reused.

lib/rubocop/cop/sorbet/signatures/setter_return_type.rb
rafaelfranca: Good point. I thought it was by sorbet already. I'll check

rafaelfranca: So I checked, Sorbet is weird about it. You can't use:

```ruby
#: -> String
attr_writter :a
```

You need to use

```ruby
#:  String
attr_writter :a
```

But that generate an equivalent sig:

```
sig { params(: String).returns(String)
```

Which correct. Now that I think about it, the right signature isn't `void` but the same type of the argument like what RBS generated, but then it would trick people that they need to return the argument, when you actually don't.

What made me write this cop is that I saw code like:

```ruby
def foo=(a)
  @foo = a
  a
end
```

And that last line is never used, because doesn't matter what you put there.

KaanOzkan: Oh right, nevermind then.

KaanOzkan (approved):

PR #390, opened 2026-07-29
Patch Lint/Void to be RBS `#: absurd` aware

rafaelfranca:
RuboCop 1.85.0+ (rubocop/rubocop#14756) started inspecting `case/when` branch bodies in `Lint/Void`. This produces false positives on RBS inline `#: absurd` type assertions — the Sorbet/RBS idiom for marking unreachable branches in exhaustive pattern matching:

```ruby
  #: ((String | Integer) x) -> void
  def foo(x)
    case x
    when String
    when Integer
    else
      x #: absurd
    end
  end
```

The `x #: absurd` is a deliberate type-level assertion that the else branch is unreachable, not a value used in void context.

Fix: prepend a Sorbet-aware behaviour onto `RuboCop::Cop::Lint::Void` (following the same pattern as `Style/MutableConstant`) that skips `check_expression` for nodes directly followed by an RBS `#: absurd` inline comment. The detection requires only whitespace between the node and the comment (so `x; y #: absurd` still flags `x`) and matches the comment text exactly (`#: absurd` with nothing but trailing whitespace), preventing both semicolon-adjacent and trailing-text false negatives.

Also updates the lockfile to rubocop 1.88.2 so the test suite exercises the `case/when` branch inspection that triggers the bug.


paracycle (approved):
Nice one!

Morriar (approved):

PR #391, opened 2026-07-29
Add new `Sorbet/TestsDefinedWithEach` cop

dduugg:
### Motivation

Sorbet only looks for test methods at the top level of a class body or `describe` block, so it can't see tests generated inside `each`. `self` stays bound to the example group rather than an instance, and calls inside the loop fail to resolve — one loop cascades into an error per call, which is usually enough to keep the file at `# typed: false`. [`test_each`/`test_each_hash`][docs] exist precisely so Sorbet can see through the loop, but nothing points you at them.

This cop finds those loops and corrects them. In a large private codebase it corrected ~1,400 loops across ~800 test files, which was enough to promote ~200 of them off `# typed: false`.

### Implementation

Fires on a block whose send is `each`, `each_pair` or `each_with_index` and whose body defines tests. Picks `test_each_hash` for hash literals and `each_pair`, materializes `each_with_index` with `to_a` so the row and index arrive as a single pair, and rewrites `|a, b|` to `|(a, b)|` — `each` yields one row per iteration for `Array`, `Hash`, and the rest of the standard library, which multiple block parameters were already auto-splatting, so that part is identity there (see the safety note below on receivers that don't).

Sorbet accepts only `it`, `before`, `after`, `describe`, `context`, `specify`, `let`, `subject`, `include_examples` and `shared_examples` at the top level of a `test_each` block; anything else raises [3507][3507]. Worth noting the error message names just the first four, so I measured the rest against Sorbet 0.6.13342 rather than reading it off the docs. Consequences:

- a loop is left alone when its body holds a statement outside that set — `it_behaves_like`, a guard clause, an assignment. It has to be every statement, not merely one accepted statement, since a single unaccepted one raises for the whole block.
- nested loops are skipped at both levels. Correcting both nests `test_each` inside `test_each`; correcting only the outer leaves a bare `each` block there; wrapping the inner one in a `describe` is rejected too. A loop chained onto another correctable loop is skipped too — correcting the outer one first would invalidate the source range the inner correction depends on, so only the inner one corrects on a given run (the outer becomes correctable on the next run, once the inner is out of the way).

Also left alone: splat, block and already-destructured parameters, block locals, a lone `each_with_index` parameter, a trailing comma that discards the rest of the row, a comment between the receiver and the selector, and — below Ruby 3.4, where wrapping it in parens isn't valid syntax — a parenthesis-less command call taking a block as the receiver.

Ships `Include`d to test and spec files only (`**/test/**/*`, `**/*_test.rb`, `**/spec/**/*`, `**/*_spec.rb`), matching the scope the body check already narrows to in practice.

Correction is `SafeAutoCorrect: false` — neither helper ships with `sorbet-runtime`, so correcting a project that hasn't defined them raises `NoMethodError`. It also assumes the receiver's `each` yields one value per iteration; a receiver whose `each` yields multiple values instead (`yield a, b`, not `yield [a, b]`) isn't detected, and correcting it silently drops everything past the first value.

### Open questions

- **Name.** `TestsDefinedWithEach` mirrors the wording of the docs section. Happy to rename.
- **Minitest.** Everything above was validated against an RSpec codebase (1,389 loops across 832 spec files); I haven't run this against a Minitest suite. The accepted top-level statements (`it`, `before`, `after`, `describe`, `context`, `specify`, `let`, `subject`, `include_examples`, `shared_examples`) line up with RSpec's DSL. `Minitest::Spec` shares much of that surface (`describe`, `it`, `before`, `after`, `let`, `subject`), but `context` and `shared_examples`/`include_examples` aren't native to it, and classic `Minitest::Test` defines tests via `def test_*` rather than a method call, which this cop's body check doesn't recognize at all — those loops would pass through uncorrected rather than getting flagged. Since [the docs][docs] are titled around Minitest, worth deciding whether this cop should extend to that DSL or stay scoped to RSpec.

[docs]: https://sorbet.org/docs/minitest#table-driven-tests-tests-defined-with-each
[3507]: https://sorbet.org/docs/error-reference#3507


rafaelfranca:
This is very much RSpec specific. We aren't users of RSpec so I'm not sure how I feel about maintaining a cop for it. Can we at least make the cop work with minitest and ActiveSupport::TestCase?

dduugg:
Fair concern, and Minitest was the open question I most wanted an answer on — I flagged it in the description precisely because I'd only validated against an RSpec suite.

For context on where the assumption came from: we found this pattern on Sorbet's own docs page for the API, which is titled ["Minitest's Spec DSL"](https://sorbet.org/docs/minitest). Its opening example subclasses `Minitest::Spec`, and all three `test_each`/`test_each_hash` examples on the page use `it` blocks. So the working assumption was that this was close to a drop-in for minitest specs too. Having dug in, that holds for one half of your question and not the other.

### The accepted-method list is Sorbet's, and Sorbet's is minitest-shaped

The cop only accepts what `test_each` accepts, and that's decided by `runUnderParameterized()` in Sorbet's `rewriter/Minitest.cc`. Of the ten methods it accepted when you looked, seven are `Minitest::Spec`'s own DSL — `it`, `before`, `after`, `let`, `subject` and `specify` are `Minitest::Spec::DSL` instance methods and `describe` is `Kernel#describe`, none of them imported from RSpec. So spec-style minitest already corrects today; the overlap with RSpec is because minitest/spec and RSpec overlap, not because the cop is aimed at RSpec.

The other three — `context`, `include_examples`, `shared_examples` — aren't minitest. Checking that turned up a real bug, which I've now fixed: the allowlist was a flat set of ten names with no arity rules, which both over- and under-shot what `test_each` takes, so the cop corrected some loops into a 3507. The arity is fixed per family, so `before(:each)`, `context "x", :aggregate` and `it "x", :slow` are all out; every accepted statement needs a block of its own with no parameters, `_1` and `it` included; `let`/`let!`/`subject` are accepted only where the loop already sits inside an example group; and the shared-example helpers take any arguments but not a splat. Sixteen names were missing, mostly the `x`/`f`-prefixed forms plus `example`, `example_group`, `focus`, `pending`, `skip`, `let!`, `include_context` and `shared_context`. All measured against Sorbet 0.6.13403, since the error message names four of the twenty-nine and no arities at all.

I re-validated that against the codebase this came from, to check the tightened allowlist wasn't just churn: of the 1,345 corrections the cop made there before, 1,342 are unchanged, 3 are dropped, and 1 is newly found. The 3 dropped are all a loop carrying metadata on an example or an example group (`it "…", :slow`, `context "…", :aggregate`), which raises 3507 — latent today only because those files are still `# typed: false`, and fatal the moment they're promoted, which is the entire point of converting them. The 1 added is a `let!` loop the old ten-name list didn't recognize.

### `ActiveSupport::TestCase`'s declarative surface can't be corrected — Sorbet rejects it

`test_each`'s body is gated by a hardcoded switch with no arm for `test`, `setup`, `teardown`, `define_method` or a plain `def`. Against Sorbet 0.6.13403:

```ruby
# typed: true
class FooTest < ActiveSupport::TestCase
  test_each([1, 2, 3]) do |x|
    test "x" do   # 3507
    end

    setup do      # 3507
    end
  end
end
```

> Only valid `it`, `before`, `after`, and `describe` blocks can appear within `test_each`

Same result with or without `--enable-experimental-rspec`, in a bare class, a `Minitest::Test`, an `ActiveSupport::TestCase`, and inside a `describe` — every host where Sorbet analyses the loop at all.

Sorbet does understand `test "..." do` — but through a different rewriter, `rewriter/TestCase.cc`, which only walks the direct statements of a `ClassDef` body and never recurses into blocks. So `test` inside `test_each` isn't merely unimplemented, it's actively rejected: teaching the cop to correct those loops would turn currently-typechecking code into a hard type error.

I measured Rails to size that. Of the 104 table-driven test-defining loops across its `**/test/**/*_test.rb`, 54 use `test "..."`, 46 use `define_method("test_...")`, and 2 use `it` (in `Arel::Spec`, which is a `Minitest::Spec`). So on a Rails-style suite this cop is close to a no-op until Sorbet's minitest rewriter grows a `test`/`setup`/`teardown` arm inside `test_each`. That seems like the right place to fix it and I'm happy to take a run at that upstream PR, but it isn't something this cop can do on its own.

### What we've attempted for Minitest in the commits here

Tests for the minitest patterns, so the boundary is pinned by the suite rather than by prose: a `Minitest::Spec` loop that does correct (modelled on Rails' `Arel::Spec` math test), plus no-offense guards for the ActiveSupport `test` macro, `setup`/`teardown`, `define_method("test_...")` and a plain `def test_x`. `Include:` already covered `**/test/**/*` and `**/*_test.rb`, so no config change was needed.

### One ask: is there a repo you'd point me at?

I went looking for a public Sorbet + minitest codebase with table-driven loops to validate against and came up empty — tapioca, packwerk, ruby-lsp, ruby-lsp-rails, identity_cache, job-iteration, maintenance_tasks, shopify-api-ruby, bootsnap, cli-ui, measured and statsd-instrument all produce zero offenses, mostly because they write `def test_*` (structurally invisible to this cop) or keep their `each` loops inside `it` blocks. Homebrew/brew is the only public `test_each` user I found, and it's RSpec. If there's a repo you'd suggest — public, or internal where you just tell me the count — I'd much rather validate against a real minitest suite than argue from the docs.

I've also flipped it to `Enabled: false` rather than `pending`, since a suite built on `ActiveSupport::TestCase` or classic `Minitest::Test` gets nothing from it — enabling it should be a deliberate choice, not something that arrives with `NewCops: enable`. That should keep it off your path entirely.

To be clear about the intent: we're offering this upstream because it seemed generally useful, not because we need it to live here. But if it's too narrow a use case for rubocop-sorbet, we're glad to keep it in [rubocop-gusto](https://github.com/Gusto/rubocop-gusto) instead — no hard feelings either way.

dduugg:
One more gap in the same class, found while tightening the allowlist and not yet fixed.

`defines_tests?` validates only the top level of the loop body, but Sorbet applies the rule recursively into nested `describe` bodies. So a loop whose top-level statement is a perfectly acceptable `describe` still corrects, even when that `describe` holds something `test_each` rejects:

```ruby
# both flagged and autocorrected today
ROWS.each do |row|
  describe row do
    doubled = row * 2      # 3507 once corrected

    it { assert(doubled) }
  end
end

OTHER.each do |row|
  describe row do
    INNER.each do |n|      # becomes a nested test_each on the second pass -> 3507
      it { assert(n) }
    end
  end
end
```

The first corrects to `test_each` → `describe` → assignment, the second to `test_each` → `describe` → `test_each`. Sorbet raises 3507 on both, measured against 0.6.13403. The nested-loop guard misses the second because `test_each` isn't one of the loop methods it looks for, so once the outer loop is corrected the inner one stops looking nested and converts on the following pass.

Neither shape appears in the codebase I validated against — zero instances across its 830 spec files — so this is a gap for other consumers rather than something the sweep hit. Happy to make the check recurse, or to document it as a known limitation next to the multi-yield `each` note, whichever you'd prefer if this lands.

KaanOzkan:
@dduugg Thanks for investigating the Minitest behavior. I checked Core and found 21 loops that define tests. All of them use `ActiveSupport::TestCase#test`; none use the spec-style methods accepted by `test_each`.

Since this cop intentionally ignores those loops, we would not be able to use or exercise it at Shopify. As Rafael said we probably shouldn't maintain it in `rubocop-sorbet` unless Sorbet first adds support for `test` inside `test_each`. Keeping it in `rubocop-gusto` seems like the better fit for now.

dduugg:
No worries, @KaanOzkan, I'll take it to another repo in that case. As always, I appreciate y'all's time in reviewing, as well as the thoughtful feedback.

PR #392, opened 2026-07-31
Drop support for Ruby 3.1 and 3.2

Morriar:
## Summary

- raise the minimum supported Ruby version to 3.3
- test Ruby 3.3, 3.4, and 4.0 in CI
- update `ruby/setup-ruby` so CI can install Ruby 4.0
- update Minitest and its exception assertion for Ruby 4.0 compatibility

## Testing

- `bin/rake test` on Ruby 4.0.6 (648 runs, 1059 assertions)
- `bin/rubocop`
- default RuboCop configuration validation
- gem build
- GitHub Actions passes on Ruby 3.3, 3.4, and 4.0


paracycle (approved):
Thanks!

paracycle:
I guess we also need to update the settings to check for the correct Ruby tests:
image

PR #393, opened 2026-08-01
Add Sorbet/StructPropName cop

### LINKED ISSUE #237: Cop suggestion: Sorbet/StructPropName for T::Struct properties
It would be great to have a cop that works like [Naming/MethodName](https://docs.rubocop.org/rubocop/cops_naming.html#namingmethodname), but for T::Struct props 

 Personally I only plan to use snake_case for this right now, but for completeness the cop could support the same options as [Naming/MethodName](https://docs.rubocop.org/rubocop/cops_naming.html#namingmethodname).

Naming/MethodName source: https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/naming/method_name.rb

Example:
```ruby
class Foo < T::Struct
  const :fooConst, String # bad
  const :foo_const, String # good
  
  prop :fooProp, T::Boolean # bad
  prop :foo_prop, T::Boolean # good
end
```

Guflly:
Adds `Sorbet/StructPropName` with the same naming styles and allow/forbid filters as `Naming/MethodName`. It covers `T::Struct` variants, `const` and `prop` calls, configuration, and nested scopes, and updates the generated cop docs.

Tests: focused cop tests, full suite, RuboCop, and documentation generators.

Closes #237.


lib/rubocop/cop/sorbet/struct_prop_name.rb
```diff
@@ -0,0 +1,101 @@
...
+        RESTRICT_ON_SEND = [:const, :prop].freeze
+
+        # @!method t_struct?(node)
+        def_node_matcher :t_struct?, <<~PATTERN
+          (const (const {nil? cbase} :T) {:Struct :ImmutableStruct :InexactStruct})
+        PATTERN
+
+        def on_send(node)
+          name_node = node.first_argument
+          return unless node.receiver.nil? && name_node&.sym_type? && within_t_struct?(node)
```
KaanOzkan: Instead of checking for `nil` receiver, should we check for `nil` or `self`?

KaanOzkan (approved):

lib/rubocop/cop/sorbet/struct_prop_name.rb
Guflly: Yes, explicit `self` should be accepted here. Updated the check and tests.

lib/rubocop/cop/sorbet/struct_prop_name.rb
```diff
@@ -0,0 +1,102 @@
...
+          name = name_node.value
+          return if matches_allowed_pattern?(name)
+
+          if forbidden_name?(name)
+            add_offense(name_node, message: format(MSG_FORBIDDEN, identifier: name))
+          else
+            check_name(node, name, name_node)
+          end
+        end
+        alias_method :on_csend, :on_send
```
Morriar: Do we really need to consider `csend` here?

Guflly: No, removed it.

KaanOzkan (approved):
Thanks

PR #395, opened 2026-08-04
Add AutocorrectStyle: rbs to Sorbet/ForbidTStruct

rafaelfranca:
ForbidTStruct autocorrects a `T::Struct` subclass into a plain class with `sig` blocks, attribute accessors, and an `initialize`. Add an `AutocorrectStyle` option (default `sig`, alternative `rbs`) so the autocorrection can emit RBS inline-comment signatures (`#:` / `#|`) instead of `sig` blocks, matching the option Sorbet/EnforceSignatures already has.

When `AutocorrectStyle: rbs`:
- attr props become `#: TYPE` comments above `attr_reader`/`attr_accessor`;
- the initializer gets a `#: (params) -> void` comment;
- `extend T::Sig` is neither inserted nor kept (an existing one is removed);
- optional keyword params (default/factory/nilable) use RBS `?name:`.

A Sorbet-to-RBS type translator maps the constructs most common on `T::Struct` props to valid RBS:
- `T.nilable(X)` -> `X?` (parenthesizing unions);
- `T.any(...)` -> ` | `, `T.all(...)` -> ` & `;
- `T::Array/Hash/Set/Range[...]` -> the bare RBS generics;
- `T::Boolean` -> `bool`, `T.untyped` -> `untyped`, `T.noreturn` -> `bot`, `T.class_of(X)` -> `singleton(X)`;
- anything unrecognized (e.g. `T.proc...`, unknown `T::X`) falls back to the valid RBS `untyped` rather than emitting malformed Sorbet syntax.

Also promote `Layout/LeadingCommentSpace: AllowRBSInlineAnnotation: true` from config/rbi.yml into the shipped default config so a normal all-cop `-A` run does not rewrite `#:` into the invalid `# :`.

Update the generated manual docs and add focused RBS autocorrect tests.

### DIFF
--- a/lib/rubocop/cop/sorbet/forbid_t_struct.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_struct.rb
@@ -35,6 +35,30 @@ module RuboCop
       #
       #     def some_method; end
       #   end
+      #
+      # @example AutocorrectStyle: rbs
+      #
+      #   # bad
+      #   class MyStruct < T::Struct
+      #     const :foo, String
+      #     prop :bar, T.nilable(Integer), default: 0
+      #   end
+      #
+      #   # good
+      #   class MyStruct
+      #     #: String
+      #     attr_reader :foo
+      #
+      #     #: Integer?
+      #     attr_accessor :bar
+      #
+      #     #: (foo: String, ?bar: Integer?) -> void
+      #     def initialize(foo:, bar: 0)
+      #       @foo = foo
+      #       @bar = bar
+      #     end
+      #   end
+      #
       class ForbidTStruct < RuboCop::Cop::Base
         include Alignment
         include RangeHelp
@@ -46,17 +70,19 @@ module RuboCop
         MSG_STRUCT = "Using `T::Struct` or its variants is deprecated in this codebase."
         MSG_PROPS = "Using `T::Props` or its variants is deprecated in this codebase."
 
+        VALID_AUTOCORRECT_STYLES = ["sig", "rbs"].freeze
         # This class walks down the class body of a T::Struct and collects all the properties that will need to be
         # translated into `attr_reader` and `attr_accessor` methods.
         class TStructWalker
           include AST::Traversal
           extend AST::NodePattern::Macros
+          attr_reader :props, :has_extend_t_sig, :extend_t_sig_node
 
-          attr_reader :props, :has_extend_t_sig
-
-          def initialize
+          def initialize(style = "sig")
             @props = []
             @has_extend_t_sig = false
+            @extend_t_sig_node = nil
+            @style = style
           end
 
           # @!method extend_t_sig?(node)
@@ -73,6 +99,7 @@ module RuboCop
             if extend_t_sig?(node)
               # So we know we won't need to generate again a `extend T::Sig` line in the new class body
               @has_extend_t_sig = true
+              @extend_t_sig_node = node
               return
             end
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/sorbet/forbid_t_struct.rb
```diff
@@ -152,10 +194,113 @@ def nilable?
...
+            sorbet_type_to_rbs(@type_node)
+          end
+
+          # Translate a Sorbet type expression AST node into RBS syntax.
+          # Handles the constructs most commonly found on `T::Struct` props:
+          # `T.nilable`, `T.any`, `T.all`, `T.untyped`, `T.class_of`, and
+          # generics like `T::Array[X]`. Any unrecognized node falls back to
+          # the valid RBS `untyped` rather than emitting malformed Sorbet
+          # syntax (e.g. `T.proc...`) into an RBS annotation.
+          def sorbet_type_to_rbs(node)
```
paracycle, line 224: Should we instead start adding an `rbi` gem dependency to this gem, so that these transformations use the existing infra we've built to do these transformations comprehensively?

rafaelfranca: Yeah. I thought about that, but I think it is too early to do that. This is the first and only cop that need to do it. When we have more, I think it would be the right call.

lib/rubocop/cop/sorbet/forbid_t_struct.rb
```diff
@@ -170,14 +315,18 @@ def on_class(node)
...
+            walker = TStructWalker.new(autocorrect_style)
             walker.walk(node.body)
 
             range = range_between(node.identifier.source_range.end_pos, node.parent_class.source_range.end_pos)
             corrector.remove(range)
             next if node.single_line?
 
-            unless walker.has_extend_t_sig
+            if rbs? && walker.extend_t_sig_node
+              corrector.remove(range_by_whole_lines(walker.extend_t_sig_node.source_range, include_final_newline: true))
```
Morriar, line 326: I think this is unsafe when the struct also has other methods with Sorbet signatures:

```ruby
class Foo < T::Sruct
  extend T::Sig

  prop :foo, Integer

  sig { void }
  def some_method; end
end
```

If we remove the `T::Sig` the `sig` call will fail if the user doesn't patch `Module` like we do.

lib/rubocop/cop/sorbet/forbid_t_struct.rb
paracycle: I think we already have the cop that requires signatures failing to offer an autocorrect in RBS mode as well. So, the way I see it, this makes the current need the second instance.

rafaelfranca: The other cop just prints `untyped` for all arguments. Let me check if bringing that dependency will simplify things or just add more code to this gem.

rafaelfranca: Here is the implementation https://github.com/Shopify/rubocop-sorbet/pull/396. I don't feel it simplified a lot, but at least closed some edge cases.

lib/rubocop/cop/sorbet/forbid_t_struct.rb
rafaelfranca: This entire autocorrect is unsafe. It is marked as such. We don't need to handle all cases.

lib/rubocop/cop/sorbet/forbid_t_struct.rb
```diff
@@ -152,10 +194,113 @@ def nilable?
...
+          def translate_t_method(method, args)
+            case method
+            when :nilable
+              inner = sorbet_type_to_rbs(args.first)
+              inner = "(#{inner})" if inner.include?(" | ") || inner.include?(" & ")
+              "#{inner}?"
+            when :any
+              args.map { |arg| sorbet_type_to_rbs(arg) }.join(" | ")
+            when :all
+              args.map { |arg| sorbet_type_to_rbs(arg) }.join(" & ")
```
Morriar, line 262: I wonder if we need to preserve grouping when `T.any` and `T.all` are nested.

For example, `T.all(T.any(A, B), C)` becomes `A | B & C`. Since RBS gives `&` higher precedence than `|`, that reads like `A | (B & C)` instead of `(A | B) & C`.

Could we parenthesize union arguments when joining an intersection, similar to what we already do for `nilable`?

rafaelfranca: Again, unsafe, so we don't need to cover all edge cases

Morriar (approved):

PR #397, opened 2026-08-06
Add autocorrect for RBS comments ↔ sig blocks signatures

Morriar:
Add an autocorrect to `Sorbet/EnforceSignatures` to use Spoom for translation between Sorbet `sig` blocks and RBS comments:

   - generate missing RBS signatures through Spoom
   - translate existing `sig` blocks to RBS with `Style: rbs`
   - translate existing RBS comments to `sig` blocks with `Style: sig`
   - support overloads, multiline continuations, accessors, and annotations
   - mark autocorrection as unsafe

## Examples

With `Style: rbs`:

```ruby
  sig { params(x: String).returns(Integer) }
  def foo(x); end
```

becomes:

```ruby
  #: (String) -> Integer
  def foo(x); end
```

With Style: sig:

```ruby
  #: (String) -> Integer
  def foo(x); end
```

becomes:

```ruby
  sig { params(x: String).returns(Integer) }
  def foo(x); end
```

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -109,7 +109,7 @@ Sorbet/EnforceSignatures:
   AutocorrectStyle: sig
   SafeAutoCorrect: false
   VersionAdded: 0.3.4
-  VersionChanged: '0.13.0'
+  VersionChanged: '<<next>>'
 
 Sorbet/EnforceSingleSigil:
   Description: 'Ensures that there is only one Sorbet sigil in a file.'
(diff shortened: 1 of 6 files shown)


config/default.yml
```diff
@@ -119,6 +107,7 @@ Sorbet/EnforceSignatures:
   Enabled: false
   Style: sig
   AutocorrectStyle: sig
+  SafeAutoCorrect: false
   VersionAdded: 0.3.4
   VersionChanged: '0.13.0'
```
rafaelfranca:
```suggestion
  VersionChanged: '<<next>>'
```

rafaelfranca (approved):

PR #400, opened 2026-08-07
Autocorrect T.bind to RBS self assertions

Morriar:
## Summary

Adds opt-in autocorrection to `Sorbet/ForbidTBind`, replacing supported `T.bind` calls with RBS inline self-type assertions.

```ruby
# Before
T.bind(self, Foo)

# After
#: self as Foo
```

```yaml
Sorbet/ForbidTBind:
  Enabled: true
  AutocorrectToRBS: true
```

It is marked unsafe because it removes the runtime `T.bind` call.

## Type translation

The cop extracts the type argument and uses `RBI::Type` to parse and render it as RBS:

```ruby
T.bind(self, T::Array[String])
```

becomes:

```ruby
#: self as Array[String]
```

This also handles Sorbet types such as `T.class_of(...)`, `T.all(...)`, and `T.untyped` through RBI's existing type translation.

`rbi ~> 0.4` is therefore added as a direct runtime dependency.

## Autocorrection constraints

The cop only autocorrects when an RBS comment can safely replace the expression:

- The first argument is `self`.
- The call is a standalone expression.
- The call ends the line.
- No RBS annotation already follows the call.
- RBI can parse the type argument.

Calls embedded in assignments, chained calls, or other expressions remain offenses without autocorrection:

```ruby
value = T.bind(self, Foo)
T.bind(self, Foo).call
consume(T.bind(self, Foo))
```

### DIFF
--- a/lib/rubocop/cop/sorbet/forbid_t_bind.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_bind.rb
@@ -17,12 +17,11 @@ module RuboCop
       #   # good
       #   #: self as Integer
       class ForbidTBind < RuboCop::Cop::Base
+        include RangeHelp
         extend AutoCorrector
 
         MSG = "Do not use `T.bind`."
-        COMMENT_START = "#"
-        HORIZONTAL_WHITESPACE = ["\t", " "].freeze
-        NEWLINES = ["\n", "\r"].freeze
+        LINE_ENDINGS = ["", "\n", "\r", "#"].freeze
         RESTRICT_ON_SEND = [:bind].freeze
 
         # @!method t_bind?(node)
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/sorbet/forbid_t_bind.rb
```diff
@@ -15,16 +17,63 @@ module Sorbet
...
+        end
+
+        def rbs_annotation_follows?(node)
+          source = processed_source.buffer.source
+          position = skip_horizontal_whitespace(source, node.source_range.end_pos)
+
+          source[position, 2] == "#:"
+        end
+
+        def skip_horizontal_whitespace(source, position)
```
paracycle: Are you sure there are no ready source navigation methods defined in RuboCop helpers to do things like this?

paracycle (approved):

lib/rubocop/cop/sorbet/forbid_t_bind.rb
Morriar: You're right, we can use `range_with_surrounding_space` to simplify a bit

PR #401, opened 2026-08-07
Extend RedundantTLetForLiteral to trailing RBS annotations

Morriar:
## What

Extend `Sorbet/RedundantTLetForLiteral` to detect redundant trailing RBS type annotations on constants.

```ruby
# bad
GREETING = "hello" #: String
PATTERN = /foo/.freeze #: Regexp
NAMES = ["alice", "bob"] #: Array[String]

# good
GREETING = "hello"
PATTERN = /foo/.freeze
NAMES = ["alice", "bob"]
```

The cop autocorrects these offenses by removing the trailing annotation.

## Why

A trailing RBS annotation such as:

```ruby
GREETING = "hello" #: String
```

is equivalent to a `T.let` declaration. When Sorbet can already infer the constant's type from its literal value, the RBS annotation is redundant for the same reason as:

```ruby
GREETING = T.let("hello", String)
```

Supporting both syntaxes keeps the cop's behavior consistent for codebases adopting RBS comments.

## Details

The RBS path follows the existing `T.let` inference rules:

- Detects matching annotations for simple literals.
- Detects matching array annotations.
- Detects frozen regexp literals on supported Sorbet versions.
- Leaves widened or mismatched annotations unchanged.
- Leaves annotations on locals, instance variables, and conditionally assigned constants unchanged.
- Preserves annotations on frozen strings and symbols because Sorbet does not infer those constant types through `.freeze`.

```ruby
# These annotations remain required
GREETING = "hello".freeze #: String
STATUS = :active.freeze #: Symbol
```

Version gating also matches the existing `T.let` behavior:

- Bare literals remain eligible regardless of the target Sorbet version.
- Frozen regexp and array inference require Sorbet `0.6.13304` or newer.

The implementation uses one parameterized offense message for both `T.let` and RBS annotations.

### DIFF
--- a/lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
+++ b/lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
@@ -5,9 +5,9 @@ require "rubocop"
 module RuboCop
   module Cop
     module Sorbet
-      # Checks for redundant `T.let` declarations where the first argument
-      # is a literal whose type Sorbet can infer automatically, so wrapping
-      # it in `T.let` is redundant.
+      # Checks for redundant `T.let` declarations and trailing RBS annotations
+      # where the assigned value is a literal whose type Sorbet can infer
+      # automatically.
       #
       # Simple literals (strings, symbols, integers, floats, regexps) infer as
       # their own class. Regexp literals are the only simple literals whose
@@ -36,6 +36,8 @@ module RuboCop
       #   STATUS = T.let(:active, Symbol)
       #   SHELLS = T.let([:bash, :zsh].freeze, T::Array[Symbol])
       #   NAMES = T.let(["alice", "bob"], T::Array[String])
+      #   RBS_GREETING = "hello" #: String
+      #   RBS_NAMES = ["alice", "bob"] #: Array[String]
       #
       #   # good
       #   MAX_RETRIES = 3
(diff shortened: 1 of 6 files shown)


lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
```diff
@@ -131,38 +134,96 @@ def on_casgn(node)
...
+
+          frozen = value_node.send_type? && value_node.method?(:freeze)
+          array_node = frozen ? value_node.receiver : value_node
+          return unless inferable_array?(array_node)
+          return unless redundant_rbs_array_annotation?(array_node, annotated_type, frozen: frozen)
+
+          register_rbs_offense(node, comment, :Array)
+        end
+
+        def trailing_rbs_annotation(node)
```
rafaelfranca: Can't we use anything in the RBS parser in this gem?

Morriar (commented):
Review fixes addressed in cb563bc.

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
Morriar: Addressed in cb563bc. I added `RBSParser.rbs_annotation_after` for trailing expression annotations, moved comment association and `#:` parsing into that shared helper, and added parser-level coverage.

rafaelfranca (approved):

PR #402, opened 2026-08-07
Reject spaced RBS comment markers

Morriar:
## What changed

Update `RBSParser` to require contiguous RBS comment markers:

- Accept `#:` and `#|`
- Reject `# :`, `#  :`, and `# |`
- Continue allowing whitespace after the marker, such as `#: () -> void` and `#|   String`

## Why

Sorbet only recognizes RBS annotations when `#` is immediately followed by `:` or `|`. The existing `\s*` matching caused RuboCop to interpret comments such as `#  : () -> String` as signatures even though Sorbet ignores them.

This aligns `RBSParser` with Sorbet’s behavior and prevents cops from acting on annotations that Sorbet does not recognize.


jesse-shopify (approved):
Nice.

PR #403, opened 2026-08-07
Autocorrect T.absurd to RBS assertions

Morriar:
Stacked on #400.

## Summary

Adds opt-in autocorrection to `Sorbet/ForbidTAbsurd`, replacing supported `T.absurd` calls with RBS inline absurd assertions.

```ruby
# Before
T.absurd(value)

# After
value #: absurd
```

```yaml
Sorbet/ForbidTAbsurd:
  Enabled: true
  AutocorrectToRBS: true
```

It is marked unsafe because it removes the runtime `T.absurd` call.

## Shared assertion correction

Extracts the correction eligibility checks shared by `ForbidTBind` and `ForbidTAbsurd` into `RBSAssertionCorrection`.

The helper:

- checks the `AutocorrectToRBS` setting
- prevents inline comments from changing surrounding Ruby syntax
- supports assignment values when the cop opts in
- reuses `RBSParser.rbs_annotation_after` to avoid overwriting an existing RBS annotation

## Autocorrection constraints

Calls nested in another expression or followed by an existing RBS annotation remain offenses without autocorrection.

### DIFF
--- a/test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
+++ b/test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
@@ -6,13 +6,11 @@ module RuboCop
   module Cop
     module Sorbet
       class ForbidTAbsurdTest < ::Minitest::Test
-        MSG = "Sorbet/ForbidTAbsurd: Do not use `T.absurd`."
-
-        def setup
-          @cop = ForbidTAbsurd.new
-        end
+        MSG = "Do not use `T.absurd`."
 
         def test_adds_offense_when_using_t_absurd
+          @cop = target_cop.new(cop_config("AutocorrectToRBS" => false))
+
           assert_offense(<<~RUBY)
             T.absurd(foo)
             ^^^^^^^^^^^^^ #{MSG}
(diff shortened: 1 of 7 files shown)


test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
```diff
@@ -6,20 +6,81 @@ module RuboCop
...
+          RUBY
+
+          assert_no_corrections
+        end
+
+        private
+
+        def target_cop
+          ForbidTAbsurd
         end
```
vinistock, line 84: It seems that in most cases we want `AutocorrectToRBS` to be `true`. How about we do something like this?


```suggestion
        def target_cop(config: { "AutocorrectToRBS" => true })
          ForbidTAbsurd.new(cop_config(config))
        end
```

vinistock (approved):

test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
Morriar: I prefer to make it explicit at this point. We can revisit if we change the default at some point.

PR #405, opened 2026-08-07
Autocorrect T.must to RBS assertions

Morriar:
## Summary

Adds opt-in autocorrection to `Sorbet/ForbidTMust`, replacing supported `T.must` calls with RBS inline non-nil assertions.

```ruby
# Before
T.must(value)

# After
value #: as !nil
```

```yaml
Sorbet/ForbidTMust:
  Enabled: true
  AutocorrectToRBS: true
```

It is marked unsafe because it removes the runtime nil check performed by `T.must`.

## Autocorrection constraints

The cop supports standalone calls and direct assignment values. Calls nested in another expression, followed by an existing RBS annotation, or nested inside another `T.must` remain offenses without autocorrection.

### DIFF
--- a/lib/rubocop/cop/sorbet/forbid_t_must.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_must.rb
@@ -35,6 +35,7 @@ module RuboCop
         private
 
         def autocorrect_t_must_to_rbs(corrector, node)
+          return if node.first_argument.splat_type?
           return if t_must?(node.first_argument)
           return unless rbs_assertion_autocorrectable?(node, allow_assignment: true)
 
--- a/lib/rubocop/cop/sorbet/forbid_t_must.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_must.rb
@@ -6,6 +6,7 @@ module RuboCop
   module Cop
     module Sorbet
       # Disallows using `T.must` anywhere.
+      # Set `AutocorrectToRBS: true` to replace supported calls with RBS inline comments.
       #
       # @example
       #
@@ -15,6 +16,9 @@ module RuboCop
       #   # good
       #   foo #: as !nil
       class ForbidTMust < RuboCop::Cop::Base
+        include RBSAssertionCorrection
+        extend AutoCorrector
+
         MSG = "Do not use `T.must`."
         RESTRICT_ON_SEND = [:must].freeze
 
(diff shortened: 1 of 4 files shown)


lib/rubocop/cop/sorbet/forbid_t_must.rb
```diff
@@ -6,6 +6,7 @@ module RuboCop
   module Cop
     module Sorbet
       # Disallows using `T.must` anywhere.
+      # Set `AutocorrectToRBS: true` to replace supported calls with RBS inline comments.
```
amomchilov, line 9: I think you should fuse this with the `not_nil!` cop

```yaml
AutocorrectTo: "RBS comment"
AutocorrectTo: ".not_nil!"
```

amomchilov: Though I suppose that's in a different gem... hmmmmm Can one gem modify another gem's cop's behaviour? I think I've seen that before with patching/prepending

Morriar: If you're talking about https://github.com/Shopify/type_toolkit/blob/main/lib/rubocop/cop/type_toolkit/prefer_not_nil.rb I think the idea is to have either `Sorbet/ForbidTMust` or `TypeToolkit/PreferNotNil` enabled but not both?

lib/rubocop/cop/sorbet/forbid_t_must.rb
```diff
@@ -15,16 +16,30 @@ module Sorbet
...
+          add_offense(node) { |corrector| autocorrect_t_must_to_rbs(corrector, node) }
         end
         alias_method :on_csend, :on_send
+
+        private
+
+        def autocorrect_t_must_to_rbs(corrector, node)
+          return if t_must?(node.first_argument)
+          return unless rbs_assertion_autocorrectable?(node, allow_assignment: true)
+
```
KaanOzkan, line 41: We should probably skip the autocorrect if the argument to `T.must` is a splat: `T.must(*arr)`. Without the function call splat can give a SyntaxError or assign different type.

Morriar: Good call, fixed 👍

KaanOzkan (approved):

PR #406, opened 2026-08-07
Autocorrect T.cast to RBS assertions

Morriar:
## Summary

Adds opt-in autocorrection to `Sorbet/ForbidTCast`, replacing supported `T.cast` calls with RBS inline type assertions.

```ruby
# Before
T.cast(value, T::Array[String])

# After
value #: as Array[String]
```

```yaml
Sorbet/ForbidTCast:
  Enabled: true
  AutocorrectToRBS: true
```

It is marked unsafe because it removes the runtime type check performed by `T.cast`. RBI translates Sorbet type expressions to RBS.

## Autocorrection constraints

The cop supports standalone calls and direct assignment values. Calls nested in another expression, nested inside another `T.cast`, followed by an existing RBS annotation, or using an untranslatable type remain offenses without autocorrection.

### DIFF
--- a/test/rubocop/cop/sorbet/forbid_t_cast_test.rb
+++ b/test/rubocop/cop/sorbet/forbid_t_cast_test.rb
@@ -6,13 +6,11 @@ module RuboCop
   module Cop
     module Sorbet
       class ForbidTCastTest < ::Minitest::Test
-        MSG = "Sorbet/ForbidTCast: Do not use `T.cast`."
-
-        def setup
-          @cop = ForbidTCast.new
-        end
+        MSG = "Do not use `T.cast`."
 
         def test_adds_offense_when_using_t_cast
+          @cop = target_cop.new(cop_config("AutocorrectToRBS" => false))
+
           assert_offense(<<~RUBY)
             T.cast(foo, String)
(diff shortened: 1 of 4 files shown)


test/rubocop/cop/sorbet/forbid_t_cast_test.rb
```diff
@@ -6,20 +6,91 @@ module RuboCop
...
+            ^^^^^^^^^^^^^^^^^ #{MSG}
+          RUBY
+
+          assert_no_corrections
+        end
+
+        private
+
+        def target_cop
+          ForbidTCast
```
amomchilov, line 106: Meh, I'd inline tbh

amomchilov (approved):

test/rubocop/cop/sorbet/forbid_t_cast_test.rb
Morriar: While I agree, we need this for `cop_config` to work in our test helper: https://github.com/Shopify/rubocop-sorbet/blob/eff3c0aa8e1f116b23d0a919dbe8e48e7848f8ab/test/test_helper.rb#L21

We _could_ change the helper but that seemed out of scope for this PR.

PR #407, opened 2026-08-07
Autocorrect T.let to RBS annotations

Morriar:
## Summary

Adds opt-in autocorrection to `Sorbet/ForbidTLet`, replacing supported `T.let` calls with RBS inline type annotations.

```ruby
# Before
T.let(value, T.nilable(T::Array[String]))

# After
value #: Array[String]?
```

```yaml
Sorbet/ForbidTLet:
  Enabled: true
  AutocorrectToRBS: true
```

It is marked unsafe because it removes the runtime type check performed by `T.let`. RBI translates Sorbet type expressions to RBS.

## Autocorrection constraints

The cop supports standalone calls and direct assignment values. Calls nested in another expression, nested inside another `T.let`, followed by an existing RBS annotation, or using an untranslatable type remain offenses without autocorrection.

### DIFF
--- a/test/rubocop/cop/sorbet/forbid_t_let_test.rb
+++ b/test/rubocop/cop/sorbet/forbid_t_let_test.rb
@@ -42,6 +42,19 @@ module RuboCop
           RUBY
         end
 
+        def test_autocorrection_preserves_a_trailing_comment
+          @cop = target_cop.new(cop_config("AutocorrectToRBS" => true))
+
+          assert_offense(<<~RUBY)
+            T.let(foo, Bar) # some comment
+            ^^^^^^^^^^^^^^^ #{MSG}
+          RUBY
+
+          assert_correction(<<~RUBY)
+            foo #: Bar # some comment
+          RUBY
+        end
+
         def test_does_not_autocorrect_t_let_nested_in_an_expression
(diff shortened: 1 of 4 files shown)


test/rubocop/cop/sorbet/forbid_t_let_test.rb
```diff
@@ -6,20 +6,91 @@ module RuboCop
...
+            ^^^^^^^^^^^^^^^^ #{MSG}
+          RUBY
+
+          assert_no_corrections
+        end
+
+        private
+
+        def target_cop
+          ForbidTLet
```
amomchilov, line 106: Same, I'd inline this

amomchilov: Or alternative, make it actually useful by taking `autocorrect_to_rbs: bool` and return the new cop object, rather than just the class

amomchilov (approved):

test/rubocop/cop/sorbet/forbid_t_let_test.rb
Morriar: same answer than https://github.com/Shopify/rubocop-sorbet/pull/406#discussion_r3749872355.

PR #409, opened 2026-08-11
Autocorrect inline else T.absurd calls

Morriar:
So this case with an inline annotation after the `else`:

```
case foo
when Foo
  ...
else T.absurd(foo)
end
```

get autocorrect to this:

```
case foo
when Foo
  ...
else foo #: absurd
end
```

### DIFF
--- a/lib/rubocop/cop/sorbet/forbid_t_absurd.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_absurd.rb
@@ -49,7 +49,7 @@ module RuboCop
 
         def inline_else_statement(node)
           parent = node.parent
-          return unless parent&.case_type?
+          return unless parent&.type?(:case, :if)
           return unless parent.children.last.equal?(node)
 
           else_range = parent.loc.else
--- a/lib/rubocop/cop/sorbet/forbid_t_absurd.rb
+++ b/lib/rubocop/cop/sorbet/forbid_t_absurd.rb
@@ -39,6 +39,22 @@ module RuboCop
 
           corrector.replace(node, "#{node.first_argument.source} #: absurd")
         end
+
+        def assertion_statement(node, allow_assignment:)
+          inline_else = inline_else_statement(node)
+          return super(inline_else, allow_assignment: allow_assignment) if inline_else
+
+          super
+        end
+
+        def inline_else_statement(node)
+          parent = node.parent
+          return unless parent&.case_type?
+          return unless parent.children.last.equal?(node)
+
+          else_range = parent.loc.else
+          parent if else_range&.source == "else" && else_range.line == node.first_line
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/rubocop/cop/sorbet/forbid_t_absurd.rb
```diff
@@ -39,6 +39,22 @@ def autocorrect_t_absurd_to_rbs(corrector, node)
...
+        def assertion_statement(node, allow_assignment:)
+          inline_else = inline_else_statement(node)
+          return super(inline_else, allow_assignment: allow_assignment) if inline_else
+
+          super
+        end
+
+        def inline_else_statement(node)
+          parent = node.parent
+          return unless parent&.case_type?
```
paracycle: I don't think checking `case` is enough:
```ruby
irb> if false
irb> then 42
irb> else 43
irb> end
#=> 43
```

Morriar: Good call, we can handle this right now 👍

test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
```diff
@@ -55,6 +55,58 @@ def test_autocorrection_preserves_a_trailing_comment
...
+
+        def test_autocorrects_t_absurd_after_an_inline_if_else
+          @cop = target_cop.new(cop_config("AutocorrectToRBS" => true))
+
+          assert_offense(<<~RUBY)
+            if false
+            then 42
+            else T.absurd(foo) # exhaustive
+                 ^^^^^^^^^^^^^ #{MSG}
+            end
```
paracycle, line 85: Can we add a test with `unless` as well please?

test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
```diff
@@ -55,6 +55,58 @@ def test_autocorrection_preserves_a_trailing_comment
...
 
+        def test_autocorrects_t_absurd_after_an_inline_else
+          @cop = target_cop.new(cop_config("AutocorrectToRBS" => true))
+
+          assert_offense(<<~RUBY)
+            case foo
+            when String then foo
+            else T.absurd(foo) # exhaustive
+                 ^^^^^^^^^^^^^ #{MSG}
+            end
```
paracycle, line 66: Should we test `case`/`in` calls as well? Or do they still not do proper type-checking anyway?

Morriar: it's still clunky: https://sorbet.run/#%23%20typed%3A%20true%0A%0Ax%20%3D%20T.let%28ARGV.first%2C%20String%29%0A%0Acase%20x%0Ain%20%2Fabc%2F%0A%20%20puts%20%22match%22%0Ain%20String%0A%20%20puts%20%22string%22%0Aelse%0A%20%20T.absurd%28x%29%0Aend

I'd pass for now.

test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
Morriar: Sorry I missed your comment, fixed 👍

paracycle (approved):

PR #410, opened 2026-08-12
Autocorrect Sorbet assertions in method arguments

Morriar:
## Summary

Extend the shared RBS assertion autocorrector to support Sorbet assertions used as method arguments.

This applies to:

- `Sorbet/ForbidTCast`
- `Sorbet/ForbidTLet`
- `Sorbet/ForbidTUnsafe`

It supports:

- Positional arguments
- Keyword arguments
- Positional splats
- Keyword splats whose asserted value is a regular expression

## Examples

```ruby
consume(first, T.unsafe(value), last)
```

is corrected to:

```ruby
consume(
  first,
  value, #: as untyped
  last
)
```

For splats:

```ruby
consume(*T.unsafe(values))
```

is corrected to:

```ruby
consume(
  *(
    values #: as untyped
  )
)
```

The parentheses are required so the RBS comment applies to the value being splatted without swallowing the closing syntax.

## Guardrails

The autocorrect declines cases where rewriting the enclosing call is ambiguous or would introduce invalid layout:

- Calls inside single-line blocks
- Calls without usable delimiters
- Already-multiline calls
- Calls containing comments
- Calls containing multiple matching assertions
- Keyword splats where `T.unsafe` receives an implicit hash

For example, this remains uncorrected:

```ruby
consume(**T.unsafe(required: value, **options))
```

Materializing that implicit hash belongs in the hash-focused follow-up.

Receiver chains are also out of scope. Core testing showed that Sorbet does not reliably apply an RBS assertion when the asserted expression continues as a receiver.

## Follow-ups

- Explicit array elements
- Explicit hash values
- Single-line block expansion
- Already-multiline calls
- Multiple assertions in one argument list

### DIFF
--- a/lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
+++ b/lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
@@ -3,15 +3,132 @@
 module RuboCop
   module Cop
     module Sorbet
-      # Shared guards for replacing Sorbet runtime assertions with RBS inline
-      # comments without changing the surrounding Ruby syntax.
+      # Shared guards and rewrites for replacing Sorbet runtime assertions
+      # with equivalent RBS inline comments.
       module RBSAssertionCorrection
         include RangeHelp
+        include Alignment
 
         COMMENT_START = "#"
         NEWLINES = ["\n", "\r"].freeze
 
         private
+
+        def autocorrect_rbs_assertion(corrector, node, allow_assignment: false)
+          context = rbs_assertion_correction_context(node, allow_assignment: allow_assignment)
+          return unless context
+
+          expression, annotation = yield
+          case context.first
+          when :direct
+            corrector.replace(node, "#{expression} #: #{annotation}")
+          when :arguments
+            autocorrect_argument_list(corrector, node, expression, annotation, context)
+          end
+        end
+
+        def rbs_assertion_correction_context(node, allow_assignment:)
+          return [:direct] if rbs_assertion_autocorrectable?(node, allow_assignment: allow_assignment)
+          return unless nested_rbs_assertion_autocorrectable?(node)
+
+          call_argument_context(node)
+        end
+
+        def call_argument_context(node)
+          parent = node.parent
+          if parent&.type?(:call) && parent.arguments.include?(node)
+            call = parent
+            target = node
+          elsif parent&.splat_type?
+            call = parent.parent
+            return unless call&.type?(:call) && call.arguments.include?(parent)
+
+            target = parent
+          elsif parent&.kwsplat_type?
+            argument = node.first_argument
+            return if argument.hash_type? && !argument.loc.begin
+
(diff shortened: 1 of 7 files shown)


lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
```diff
@@ -3,16 +3,132 @@
...
+          end
+        end
+
+        def one_matching_assertion?(container, node)
+          container.each_node(:send).count do |candidate|
+            candidate.method?(node.method_name) &&
+              candidate.receiver&.const_type? &&
+              candidate.receiver.const_name == "T"
+          end == 1
+        end
```
KaanOzkan, line 118: Looks like we don't want to autocorrect when we have multiple assertions but I think this returns true for multiple **separate** assertions, e.g. `foo(T.let(...), T.cast(...))`. Should we skip those too?

lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
```diff
@@ -3,16 +3,132 @@
...
+          _, container, items, target = context
+          indentation = container.source_range.source_line[/\A[ \t]*/]
+          item_indentation = indentation + (" " * configured_indentation_width)
+          splat = target.type?(:splat, :kwsplat)
+          target_index = items.index(target)
+          if splat
+            nested_indentation = item_indentation + (" " * configured_indentation_width)
+            replacement = "(\n#{nested_indentation}#{replacement}\n#{item_indentation})"
+          elsif target_index < items.length - 1
+            replacement = replacement.sub(" #:", ", #:")
```
Morriar: This can rewrite inside the expression source instead of the RBS marker. For example, `consume(first, T.unsafe("already #: marker"), last)` currently autocorrects to:

```ruby
consume(
  first,
  "already, #: marker" #: as untyped
  last
)
```

That both changes the string literal and drops the argument separator. Could we target the last ` #:`, or otherwise have the caller pass the expression and annotation separately, so the comma is inserted next to the generated RBS comment?

lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
Morriar: Good catch. I think this is fine for now: the guard is scoped to the current cop, and these migrations run one assertion cop at a time. We can broaden it to all supported assertion methods if we need combined runs later.

lib/rubocop/cop/sorbet/mixin/rbs_assertion_correction.rb
Morriar: Fixed in 255b396. The correction block now returns the expression and annotation separately, so comma insertion is built next to the generated RBS marker and cannot rewrite a `#:` sequence inside the expression. Added the string-literal regression case from this thread.

KaanOzkan (approved):

PR #411, opened 2026-08-13
Autocorrect RBS assertions on receivers

Morriar:
## Summary

Extend the shared RBS assertion autocorrector to support Sorbet assertions used as method receivers.

This applies to:

- `Sorbet/ForbidTCast`
- `Sorbet/ForbidTLet`
- `Sorbet/ForbidTUnsafe`

This PR is stacked on #410, which adds support for assertions used as method arguments.

## Examples

A receiver chain beginning a statement:

```ruby
T.unsafe(foo).bar
```

is corrected to:

```ruby
foo #: as untyped
  .bar
```

The same applies to longer chains and safe navigation:

```ruby
T.cast(foo, String).bar.baz
T.let(foo, String)&.bar
```

become:

```ruby
foo #: as String
  .bar.baz

foo #: String
  &.bar
```

Receiver chains used as the complete right-hand side of an assignment are also supported:

```ruby
x = T.unsafe(foo).bar
```

becomes:

```ruby
x = foo #: as untyped
  .bar
```

## Semantics

The generated RBS comments preserve the original Sorbet operation:

- `T.unsafe(foo)` → `foo #: as untyped`
- `T.cast(foo, Type)` → `foo #: as Type`
- `T.let(foo, Type)` → `foo #: Type`

In particular, `T.unsafe` uses `#: as untyped`, not `#: untyped`. The latter represents `T.let(foo, T.untyped)` and would change semantics.

The generated forms were validated against Sorbet’s RBS rewrite phase.

## Guardrails

The autocorrect declines receiver chains when rewriting only the receiver would cause the trailing RBS comment to consume surrounding syntax.

Unsupported embedded contexts include:

```ruby
wrapper(T.unsafe(foo).bar)
return T.unsafe(foo).bar
other + T.unsafe(foo).bar
```

These require restructuring the enclosing expression, generally by adding parentheses and expanding it across multiple lines.

The correction also declines:

- Receiver chains inside single-line blocks
- Multiline assertions or receiver chains
- Receiver chains containing comments
- Receiver chains containing multiple matching assertions
- Receiver chains with an existing RBS annotation

`T.bind` and `T.absurd` are intentionally excluded. Their RBS replacements are statement-oriented and do not compose naturally as receiver expressions.


test/rubocop/cop/sorbet/forbid_t_unsafe_test.rb
```diff
@@ -55,6 +55,59 @@ def test_autocorrection_preserves_a_trailing_comment
           RUBY
         end
 
+        def test_autocorrects_t_unsafe_used_as_a_receiver
```
vinistock, line 58: Can we add an example where the T unsafe happens in the middle of a call chain? For example:

```ruby
T.unsafe(foo.bar).baz

# Needs to become
foo.bar #: as unsafe
  .baz
```

vinistock (approved):

PR #412, opened 2026-08-13
Add `ForbidSorbetGenericTypesInRBSComments` cop

dejmedus:
Generic types (`Array`, `Hash`, etc) in RBS comments don't need to be [`T::` Sorbet types](https://sorbet.org/docs/rbs-support#special-behaviors)

This cop will display `Sorbet/ForbidSorbetGenericTypesInRBSComments: Use <type> instead of Sorbet's T::<type> in RBS comments` and optionally unsafely autocorrect:

``` diff
- #: (String name, T::Array[Symbol] arr) -> void
+ #: (String name, Array[Symbol] arr) -> void
```

<details><summary>Demo</summary>

https://github.com/user-attachments/assets/757ee0cb-d225-4696-803a-768779994e67

</details>


config/default.yml
```diff
@@ -5,6 +5,12 @@
 Layout/LeadingCommentSpace:
   AllowRBSInlineAnnotation: true
 
+Sorbet/ForbidSorbetGenericTypesInRBSComments:
```
dejmedus, line 8: Is this name too long? 🤔

lib/rubocop/cop/sorbet/forbid_sorbet_generic_types_in_rbs_comments.rb
```diff
@@ -0,0 +1,65 @@
...
+          (?<rbs_type>
+            Enumerator::(?:Chain|Lazy)|
+            Array|
+            Class|
+            Enumerable|
+            Enumerator|
+            Hash|
+            Module|
+            Set|
+            Range
```
paracycle: Can we keep this list alphabetical and explicit?

```suggestion
            Array|
            Class|
            Enumerable|
            Enumerator|
            Enumerator::Chain|
            Enumerator::Lazy|
            Hash|
            Module|
            Range
            Set|
```

lib/rubocop/cop/sorbet/forbid_sorbet_generic_types_in_rbs_comments.rb
```diff
@@ -0,0 +1,65 @@
...
+      #   #: (T::Array[String]) -> ::T::Hash[Symbol, Integer]
+      #
+      #   # good
+      #   #: (Array[String]) -> Hash[Symbol, Integer]
+      class ForbidSorbetGenericTypesInRBSComments < Base
+        extend AutoCorrector
+
+        MSG = "Use `%<rbs_type>s` instead of Sorbet's `%<sorbet_type>s` in RBS comments."
+
+        SORBET_GENERIC_TYPE = %r{
```
paracycle: Can we not actually parse the RBS signature and act on the types in there, instead of attacking this with a regular expression?

dejmedus: A bit trickier as Rubocop sees comments as text, but there are some [RBS parsing helpers](https://github.com/Shopify/rubocop-sorbet/blob/main/lib/rubocop/sorbet/rbs_parser.rb) already. I tried out a few solutions and I'm leaning towards using [RBS lexer to grab comment tokens](https://github.com/Shopify/rubocop-sorbet/pull/412/commits/a658dde0bef080cd73891a4dba93eb55af6a8bc5) and then check if they match the pattern

Though I can't decide if I prefer the added complexity to the [regex](https://github.com/Shopify/rubocop-sorbet/pull/412/commits/15249672fb71543bdb2f663e9efcdaaae4c57df5). What do you think?

lib/rubocop/cop/sorbet/forbid_sorbet_generic_types_in_rbs_comments.rb
dejmedus: Updated:)

lib/rubocop/sorbet/rbs_parser.rb
```diff
@@ -49,6 +55,23 @@ def rbs_annotation_after(processed_source, node)
           [comment, annotation]
         end
 
+        # True if `comment` begins with a `#:` a `#|`
+        def rbs_comment?(comment)
+          comment.text.match?(RBS_SIGNATURE_PREFIX) ||
+            comment.text.match?(RBS_CONTINUATION_PREFIX)
```
Morriar: Should we make continuation detection group-aware before using it for this cop? `rbs_signatures_before` intentionally drops orphan `#|` lines, but this helper now makes the new cop scan every `#|` comment independently. That means a regular comment like `#| mentions T::Array[String]` gets autocorrected even though it isn't an RBS continuation.

Let's either only process `#|` when it is contiguous after a `#:`/`#|` RBS line, or add a regression test if we do want orphan continuations to be treated as RBS comments.

lib/rubocop/cop/sorbet/forbid_sorbet_generic_types_in_rbs_comments.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            sorbet_type = type_tokens[...-1] # drops `[`
+            yield offense_range(comment, sorbet_type), rbs_type
+          end
+        end
+
+        # Group tokens into types, keeping the opening bracket
+        # when the type has arguments `[ `T`, `::`, `Array`, `[` ]`
+        def types(comment)
+          ::RuboCop::Sorbet::RBSParser.rbs_tokens(comment).chunk_while do |left, right|
```
Morriar, line 70: It may be cleaner to reuse the actual RBS parser, take a look at how we do it in Spoom: https://github.com/Shopify/rbi/blob/27c061f31d93a4d5a625b8eb915f8d3e1d2a6c1a/lib/rbi/rewriters/translate_rbs_sigs.rb#L60-L61.

Morriar: Also I wonder if we could reuse https://github.com/Shopify/rbi/blob/27c061f31d93a4d5a625b8eb915f8d3e1d2a6c1a/lib/rbi/rbs/type_translator.rb#L220.

dejmedus: I tried that out but found using a [regex](https://github.com/Shopify/rubocop-sorbet/pull/412/changes/15249672fb71543bdb2f663e9efcdaaae4c57df5#diff-0721b546640b77ce12e55c78ab9f36e49dd75198390ed49ad7e7321ae624521eR37) or the [Lexer](https://github.com/Shopify/rubocop-sorbet/pull/412/commits/f835aabd32ae520e604d177906011dc78d0fe286) buys us not having to ensure comments are in a parsable state (multilines are all together, etc) or differentiate between signatures, type aliases, and annotations/assertions the way Spoom has to. Instead we can just loop comments and look for an RBS marker + anything that matches `T::{generic type}`. Wdyt? 

> Also I wonder if we could reuse

Great point, thank you! Regardless of the solution we go with I can swap out the hardcoded list for `RUNTIME_GENERIC_TYPES`. I believe reusing the helper itself would require a translator instance though, so maybe "cheaper" to let the cop handle converting

Morriar: What happens with a signature like this one:

```rb
#: -> void # some comment with T::Array
def foo; end
```

I'm guessing we're also touching the comment after the sig?

KaanOzkan: No the lexer tokenizes everything after the second `#` as a trailing comment, so `T::Array` is not part of the RBS type token we look at. I added a test to codify it.

PR #413, opened 2026-08-19
Update ForbidTAbsurd autocorrection for new RBS syntax

Morriar:
> [!IMPORTANT]
> This PR depends on [Shopify/sorbet#871](https://github.com/Shopify/sorbet/pull/871). That PR must be merged and released first so `0.6.99999` can be replaced with the actual `sorbet-static` version before this PR merges.

## Summary

Updates the opt-in `Sorbet/ForbidTAbsurd` autocorrection to support the new RBS `absurd` syntax introduced by [Shopify/sorbet#871](https://github.com/Shopify/sorbet/pull/871).

The correction depends on the project's `sorbet-static` version.

Before the new syntax is available:

```ruby
# Before
T.absurd(value)

# After
value #: absurd
```

With the new syntax:

```ruby
# Before
T.absurd(value)

# After
raise #: absurd(value)
```

`AutocorrectToRBS` remains disabled by default:

```yaml
Sorbet/ForbidTAbsurd:
  Enabled: true
  AutocorrectToRBS: true
```

## Version gate

The new correction is gated using `TargetSorbetVersion`.

For now, `0.6.99999` is used as a placeholder cutoff:

- `< 0.6.99999`: use the existing `value #: absurd` correction
- `>= 0.6.99999`: use `raise #: absurd(value)`

The placeholder must be replaced with the actual `sorbet-static` release containing Shopify/sorbet#871.

## New-syntax constraints

The new RBS syntax requires `absurd` to annotate a standalone `raise` and explicitly name the exhaustiveness subject.

Autocorrection is therefore limited to operands supported by Sorbet:

- local variables
- instance variables
- class variables
- global variables
- `self`

Calls with constants, method calls, or other expressions remain offenses without autocorrection.

Assignment contexts are also left uncorrected with the new syntax:

```ruby
result = T.absurd(value)
```

Replacing only the call would make the trailing RBS annotation bind incorrectly. The legacy correction continues to support assignment contexts.

### DIFF
--- a/test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
+++ b/test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
@@ -7,6 +7,11 @@ module RuboCop
     module Sorbet
       class ForbidTAbsurdTest < ::Minitest::Test
         MSG = "Do not use `T.absurd`."
+        DUMMY_RBS_ABSURD_VERSION = ForbidTAbsurd::MINIMUM_RBS_ABSURD_VERSION
+
+        def setup
+          stub_sorbet_static_version("0.6.99998")
+        end
 
         def test_adds_offense_when_using_t_absurd
           @cop = target_cop.new(cop_config("AutocorrectToRBS" => false))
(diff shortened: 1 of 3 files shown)


test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
```diff
@@ -42,6 +47,84 @@ def test_autocorrects_t_absurd_to_an_rbs_absurd_assertion_when_enabled
...
+          assert_offense(<<~RUBY)
+            def check(foo)
+              T.absurd(foo)
+              ^^^^^^^^^^^^^ #{MSG}
+            end
+          RUBY
+
+          assert_correction(<<~RUBY)
+            def check(foo)
+              raise #: absurd(foo)
```
vinistock, line 63: Should we raise a specific error or include an error message? If we ever wanted to understand how many times we hit an `absurd` in production, it will be difficult to filter without a specific error class or message.

vinistock (approved):

test/rubocop/cop/sorbet/forbid_t_absurd_test.rb
Morriar: I couldn't figure the right built-in error for this. What do you think of `TypeError.new("Control flow reached T.absurd.")`?

Morriar: Also note that it's going to be hard to enforce outside of the mass rewrite

vinistock: Oh...yeah, if this is not enforced by Sorbet, then it will indeed be a best effort. 

I like your `TypeError` suggestion. Do you think it's worth proposing that Sorbet enforces that exact form?

I can also see arguments against it. Imagine a product team building something and they want to raise their own error (or even just customize the message with some extra information).

PR #414, opened 2026-08-20
Extend redundant literal checks to local variables

Morriar:
## Summary

Extend `Sorbet/RedundantTLetForLiteral` to detect redundant literal annotations on local variables.

The cop now handles both `T.let` and RBS inline annotations:

```ruby
# bad
count = T.let(1, Integer)
name = "" #: String
rate = 1.0 #: Float

# good
count = 1
name = ""
rate = 1.0
```

It also detects exact boolean and nil types:

```ruby
# bad
enabled = T.let(true, TrueClass)
disabled = false #: false
value = nil #: nil
```

Broader annotations remain valid because they widen the inferred singleton type:

```ruby
# good
enabled = T.let(true, T::Boolean)
disabled = false #: bool
value = nil #: String?
```

When a value from an untyped source is assigned inside a block, cast that value at the reassignment boundary rather than annotating the initial literal:

```ruby
# bad
observed_locale = T.let("", String)

records.each do |record|
  observed_locale = record.locale
end

# good
observed_locale = ""

records.each do |record|
  observed_locale = record.locale #: as String
end
```

### DIFF
--- a/lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
+++ b/lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
@@ -14,6 +14,10 @@ module RuboCop
       # inference survives a `.freeze` call (Sorbet 0.6.13304+), so
       # `T.let(/foo/.freeze, Regexp)` is also redundant; other frozen simple
       # literals (e.g. `"hello".freeze`) are not inferred and still need `T.let`.
+      # Exact annotations on `true`, `false`, and `nil` are also redundant,
+      # whether expressed as `T.let(..., TrueClass)` or RBS singleton types.
+      # With RBS comments enabled, Sorbet infers the same `TrueClass`,
+      # `FalseClass`, and `NilClass` types without them.
       #
       # Array literals of simple literals are also inferred:
       #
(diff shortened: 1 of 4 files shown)


KaanOzkan (commented):
There's value in a T.let such as [this](https://sorbet.run/#%23%20typed%3A%20true%0A%0Ax%20%3D%20T.let%280%2C%20Integer%29%0Ax%20%3D%20%22bad%22%0A%0Ay%20%3D%200%0Ay%20%3D%20%22bad%22%20%23%20does%20not%20error) that errors with this current version. Wdyt of only erroring when there **isn't** a reassignment?

Although I'm not sure if supporting this use case is worth the loss of reassignment cases you caught currently.

lib/rubocop/cop/sorbet/redundant_t_let_for_literal.rb
```diff
@@ -144,7 +182,19 @@ def on_casgn(node)
             end
           end
 
-          check_rbs_annotation(node)
+          check_rbs_annotation(node, include_arrays: true)
+        end
+
+        def on_lvasgn(node)
```
Morriar: I think we need to avoid flagging local annotations once the local is reassigned.

`T.let` and trailing RBS annotations don't just duplicate the initializer's inferred type. They pin the variable type for later writes:

```ruby
x = T.let("", String)
x = 1 # Sorbet errors

y = ""
y = 1 # no error
```

So this can create false positives for locals used as accumulators or assigned after an untyped boundary. Could we either keep this scoped to constants, or skip locals when the same variable is assigned again in the local scope?

PR #416, opened 2026-09-02
Enable AllowRBSInlineAnnotation for Lint/SelfAssignment by default

rafaelfranca:
When using rubocop-sorbet we should allow RBS inline annotations in all cops that have that option.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -5,6 +5,9 @@
 Layout/LeadingCommentSpace:
   AllowRBSInlineAnnotation: true
 
+Lint/SelfAssignment:
+  AllowRBSInlineAnnotation: true
+
 Sorbet/ForbidTHelpers:
   Description: 'Forbid usage of T::Helpers via `extend` or `include`.'
   Enabled: false


Morriar (approved):

config/default.yml
```diff
@@ -5,6 +5,9 @@
 Layout/LeadingCommentSpace:
   AllowRBSInlineAnnotation: true
 
+Lint/SelfAssignment:
+  AllowRBSInlineAnnotation: true
```
Morriar: Let's bump the runtime RuboCop dependency too. `AllowRBSInlineAnnotation` for `Lint/SelfAssignment` is only supported starting in RuboCop 1.77.0, but `rubocop-sorbet` still allows `rubocop >= 1.75.2`. With 1.75.x/1.76.x, loading this config warns:

```text
Lint/SelfAssignment does not support AllowRBSInlineAnnotation parameter.
```

Otherwise users pinned to a still-valid dependency range get an unsupported config.

```ruby
spec.add_runtime_dependency("rubocop", ">= 1.77.0")
```

PR #396, opened 2026-08-04
Use RBI and RBS gems to parse and generate RBS

rafaelfranca:

PR #417, opened 2026-09-02
Use RBI and RBS gems to parse and generate RBS

rafaelfranca:
## Summary

- Use RBI for Sorbet property type conversion in `Sorbet/ForbidTStruct`.
- Use RBI for new RBS signature suggestions, including unusual Ruby parameter forms.
- Use RBS parser locations for setter return corrections.
- Preserve the current Spoom signature conversion path from #397.

Supersedes #396.

## Test plan

- `bundle exec ruby -Itest test/rubocop/cop/sorbet/forbid_t_struct_test.rb`
- `bundle exec ruby -Itest test/rubocop/cop/sorbet/signatures/enforce_signatures_test.rb`
- `bundle exec ruby -Itest test/rubocop/sorbet/rbs_parser_test.rb`
- `bundle exec ruby -Itest test/rubocop/cop/sorbet/signatures/setter_return_type_test.rb`
- `bin/rake test`

### DIFF
--- a/test/rubocop/cop/sorbet/signatures/setter_return_type_test.rb
+++ b/test/rubocop/cop/sorbet/signatures/setter_return_type_test.rb
@@ -297,14 +297,20 @@ module RuboCop
           def test_offense_for_rbs_multiline_union_return
             assert_offense(<<~RUBY)
               #: (String) ->
-              #| Integer |
-                 ^^^^^^^ #{MSG}
-              #| String
+              #| (Integer | String)
+                  ^^^^^^^ #{MSG}
               def name=(name); end
             RUBY
             assert_correction(<<~RUBY)
               #: (String) ->
-              #| void
+              #| (void)
+              def name=(name); end
+            RUBY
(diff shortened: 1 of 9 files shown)


test/rubocop/cop/sorbet/signatures/setter_return_type_test.rb
```diff
@@ -297,19 +297,25 @@ def name=(name); end
...
             assert_correction(<<~RUBY)
               #: (String) ->
-              #| void
+              #| (void)
               def name=(name); end
             RUBY
           end
 
           def test_no_offense_for_sig_with_void
+          def test_no_offense_for_rbs_unparenthesized_union_return
```
Morriar: This new test is nested under `test_no_offense_for_sig_with_void`, so Minitest does not discover or run it. It also trips `Lint/NestedMethodDefinition` in the lint job. Let's add the missing `end` before this method so the new case runs as its own test.

